PR #1, opened 2020-05-12
Use edge RuboCop for `internal_investigation` task

koic:
This configuration reproduces `internal_investigation` task of rubocop-hq/rubocop repo.

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -38,7 +38,6 @@ jobs:
       - name: install dependencies
         run: |
           bundle install --jobs 3 --retry 3
-          gem install rubocop rubocop-performance rubocop-rspec
       - name: spec
         run:  bundle exec rake spec
       - name: internal_investigation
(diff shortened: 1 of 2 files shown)


marcandre:
Thanks for checking this out.

I feel it's best *not* to add `rubocop` as a dependency to be sure that `rubocop-ast` is well self-containted.

Bundler setups all dependencies so it would be hard to detect if the `rubocop-ast` gem was actually requiring some code from the main rubocop gem. Specs would pass, but the gem wouldn't actually be self contained and anyone requiring just this gem could have their code crash...

koic:
Ah... sure! rubocop-ast will be a widely used gem outside the RoboCop core. I misunderstood this gem as part of RuboCop 💦 Thank you for gemify and explanation!

marcandre:
😆  Not sure it will be *widely* used outside of RuboCop, but the goal is that it can be used separately 😸 

I will add a note in the `gemspec` and `Gemfile`.

PR #13, opened 2020-05-26
[Fix #12] SendNode#attribute_accessor? requires at least one argument.

### LINKED ISSUE #12: attribute_accessor? is too eager
It will capture calls without parameters:

```ruby
RuboCop::ProcessedSource.new('do_something_with attr, asdf', RUBY_VERSION.to_f).ast.children[2].attribute_accessor? # => [:attr, []]
```

I suspect the pattern should be changed to something like:

```
(send nil? ${:attr_reader :attr_writer :attr_accessor :attr} $({sym str} _)+)
```

marcandre:


lib/rubocop/ast/node/send_node.rb
```diff
@@ -10,7 +10,8 @@ class SendNode < Node
       include MethodDispatchNode
 
       def_node_matcher :attribute_accessor?, <<~PATTERN
-        (send nil? ${:attr_reader :attr_writer :attr_accessor :attr} $...)
+        [(send nil? ${:attr_reader :attr_writer :attr_accessor :attr} $...)
+         (_    _    _                                                 _ ...)]
```
Drenmi, line 14: Should we have an issue that collects all the missing constructs in NodePattern?

In this case seems like we're missing the ability to match "one or more of the following".

spec/rubocop/ast/pair_node_spec.rb
```diff
@@ -37,20 +37,6 @@
...
-
-      it { expect(pair_node.colon?).to be_falsey }
-    end
-
-    context 'when using a colon delimiter' do
-      let(:source) { '{ a: 1 }' }
-
-      it { expect(pair_node.colon?).to be_truthy }
-    end
-  end
```
Drenmi, line 52: These seem unrelated to the change. 🤔

marcandre: You are right, although it was needed to have the CI pass. Note that it's a separate commit. Seemed easier than create a PR just for that, wait for CI, merge it, rebase this one, push, wait for CI, merge 😆 . Let me know if it causes issues

lib/rubocop/ast/node/send_node.rb
marcandre: Yes, good point. I wonder how often it actually happens. I was thinking that performance-wise it was basically not an issue, but clearly this is kind of obscure.

I was also wondering how useful it was to return `[:attr, an_array_of_nodes]` instead of `[:attr, *an_array_of_nodes]`. I'm still unclear as to how much compatibility we want to keep between releases...

PR #16, opened 2020-05-31
Fix rubocop offense for redundant regex escape

tejasbubane:
Example failure: https://github.com/rubocop-hq/rubocop-ast/pull/15/checks?check_run_id=724719610


bbatsov:
Thanks!

PR #19, opened 2020-06-01
Fix regex offense reported by rubocop

tejasbubane:
Example failure: https://github.com/rubocop-hq/rubocop-ast/pull/18/checks?check_run_id=728073363


koic:
Thanks!

Issue #5, opened 2020-05-12
CI should check different version of rubocop


marcandre:
It should check against earliest compatible rubocop as well as the `master` branch

marcandre:
We also need to run the full RuboCop test suite

marcandre:
The full suite is now run on RuboCop 0.84 too.

PR #20, opened 2020-06-04
Add RegexpNode regopt predicates

owst:
As suggested by @bbatsov here:
  https://github.com/rubocop-hq/rubocop/pull/8073#discussion_r433131081

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New features
 
 * [#4](https://github.com/rubocop-hq/rubocop-ast/issues/4): Add `interpolation?` for `RegexpNode`. ([@tejasbubane][])
+* [#20](https://github.com/rubocop-hq/rubocop-ast/pull/20): Add option predicates for `RegexpNode`. ([@owst][])
 
 ## 0.0.3 (2020-05-15)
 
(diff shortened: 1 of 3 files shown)


marcandre:
Looks great!
Would you also add the "o" option please?

owst:
Oops - accidental omission, thanks @marcandre !

I've never used the `o` option before, and I struggled to find an example where it causes a change in behaviour. I finally came up with:
```
[1] pry(main)> Array.new(3) { |i| /#{i}/ }
=> [/0/, /1/, /2/]
[2] pry(main)> Array.new(3) { |i| /#{i}/o }
=> [/0/, /0/, /0/]
```
even then, I'm not sure when this would be useful! 

The Perl [docs say](https://perldoc.perl.org/perlre.html#Other-Modifiers):
>   o  - pretend to optimize your code, but actually introduce bugs

😄

bbatsov:
Thanks!

marcandre:
Oh, that actually breaks `RuboCop`. Probably a naming conflict with `multiline?`. I'm reopening #5

owst:
Good catch @marcandre and 👍 for running RuboCop specs too

PR #30, opened 2020-06-10
Add struct_constructor?, class_definition? and module_definition? matchers

### LINKED ISSUE #28: Missing patterns for Struct
We need a `struct_constructor?` matcher (see #8122)
A more general "class_definition?` matcher that would match `class Foo < Bar`, `Struct.new(...) do` and `Class.new(...) do` would be nice too

tejasbubane:
Closes #28

### DIFF
--- a/lib/rubocop/ast/node.rb
+++ b/lib/rubocop/ast/node.rb
@@ -510,6 +510,21 @@ module RuboCop
          (block (send #global_const?({:Class :Module}) :new ...) ...)}
       PATTERN
 
+      def_node_matcher :struct_constructor?, <<~PATTERN
+        (block (send #global_const?(:Struct) :new ...) _ $_)
+      PATTERN
+
+      def_node_matcher :class_definition?, <<~PATTERN
+        {(class _ _ $_)
+         (sclass _ $_)
+         (block (send #global_const?({:Struct :Class}) :new ...) _ $_)}
+      PATTERN
+
+      def_node_matcher :module_definition?, <<~PATTERN
+        {(module _ $_)
+         (block (send #global_const?(:Module) :new ...) _ $_)}
+      PATTERN
+
       # Some expressions are evaluated for their value, some for their side
       # effects, and some for both
       # If we know that an expression is useful only for its side effects, that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * [#55](https://github.com/rubocop-hq/rubocop-ast/pull/55): Add `ProcessedSource#line_with_comment?`. ([@marcandre][])
 * [#63](https://github.com/rubocop-hq/rubocop-ast/pull/63): NodePattern now supports patterns as arguments to predicate and functions. ([@marcandre][])
 * [#64](https://github.com/rubocop-hq/rubocop-ast/pull/64): Add `Node#global_const?`. ([@marcandre][])
+* [#28](https://github.com/rubocop-hq/rubocop-ast/issues/28): Add `struct_constructor?`, `class_definition?` and `module_definition?` matchers. ([@tejasbubane][])
 
 ### Bug fixes
 
--- a/spec/rubocop/ast/node_spec.rb
+++ b/spec/rubocop/ast/node_spec.rb
@@ -373,4 +373,273 @@ RSpec.describe RuboCop::AST::Node do
       end
     end
   end
+
+  describe '#class_constructor?' do
+    context 'class definition with a block' do
+      let(:src) { 'Class.new { a = 42 }' }
+
+      it 'matches' do
+        expect(node.class_constructor?).to eq(true)
+      end
+    end
+
+    context 'module definition with a block' do
+      let(:src) { 'Module.new { a = 42 }' }
+
+      it 'matches' do
+        expect(node.class_constructor?).to eq(true)
+      end
+    end
+
+    context 'class definition' do
+      let(:src) { 'class Foo; a = 42; end' }
+
+      it 'does not match' do
+        expect(node.class_constructor?).to eq(nil)
+      end
+    end
+
+    context 'class definition on outer scope' do
+      let(:src) { '::Class.new { a = 42 }' }
+
+      it 'matches' do
+        expect(node.class_constructor?).to eq(true)
+      end
+    end
+  end
+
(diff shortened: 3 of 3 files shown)


lib/rubocop/ast/node.rb
```diff
@@ -497,6 +497,20 @@ def guard_clause?
          (block (send (const nil? {:Class :Module}) :new ...) ...)}
       PATTERN
 
+      def_node_matcher :struct_constructor?, <<~PATTERN
+        (block (send (const nil? :Struct) :new ...) ...)
+      PATTERN
+
+      def_node_matcher :class_definition?, <<~PATTERN
+        {(class (const nil? _) {nil? (const nil? _)} ...)
```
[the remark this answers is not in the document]
marcandre: ping @tejasbubane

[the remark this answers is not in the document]
tejasbubane: Done

lib/rubocop/ast/node.rb
```diff
@@ -497,6 +497,21 @@ def guard_clause?
...
+
+      def_node_matcher :class_definition?, <<~PATTERN
+        {(class (const _ _) ...)
+         (casgn nil? _ (block (send (const nil? {:Struct :Class}) :new ...) ...))
+         (sclass {(self) (send nil? _)} ...)}
+      PATTERN
+
+      def_node_matcher :module_definition?, <<~PATTERN
+        {(module (const _ _) ...)
+         (casgn nil? _ (block (send (const nil? :Module) :new ...) ...))}
```
marcandre, line 512: And here...

lib/rubocop/ast/node.rb
```diff
@@ -508,6 +508,21 @@ def guard_clause?
          (block (send (const nil? {:Class :Module}) :new ...) ...)}
       PATTERN
 
+      def_node_matcher :struct_constructor?, <<~PATTERN
+        (block (send (const {nil? (cbase)} :Struct) :new ...) ...)
+      PATTERN
+
+      def_node_matcher :class_definition?, <<~PATTERN
+        {(class (const _ _) ...)
```
marcandre, line 516: I would simplify it to `(class ...)` (or am I missing something?)

lib/rubocop/ast/node.rb
```diff
@@ -508,6 +508,21 @@ def guard_clause?
...
+         (casgn nil? _ (block (send (const {nil? (cbase)} {:Struct :Class}) :new ...) ...))
+         (sclass {(self) (send nil? _)} ...)}
```
marcandre, line 518: Same here... The only point in being precise is to exclude something. There are no `sclass` to exclude, and by being more precise you create a pattern that will be slower (for nothing) or wrong (here you are excluding `class << obj` by mistake)

lib/rubocop/ast/node.rb
```diff
@@ -508,6 +508,21 @@ def guard_clause?
...
+      PATTERN
+
+      def_node_matcher :module_definition?, <<~PATTERN
+        {(module (const _ _) ...)
+         (casgn nil? _ (block (send (const {nil? (cbase)} :Module) :new ...) ...))}
```
marcandre, line 523: The `casgn` is wrong here... Too precise (excluding `Foo::Bar = Module.new do`) and not actually relevant. `include(Module.new do ....` would be strange but should be included.
Be more lazy: be as non-specific as you can 😄

marcandre:
Also, I think `class_constructor?` has an issue with `::Class.new`.
I'll reiterate that I think a `top_level_constant` would probably be useful, it seems an easy error to make

bbatsov:
@tejasbubane ping :-)

tejasbubane:
@marcandre I have pushed some corrections & additional tests. I gave a shot at `top_level_constant` but could not wrap my head around different argument sizes. Here's what I tried so far:

```ruby
def_node_matcher :top_level_constant?, '(const {nil? (cbase)} {%1 %2})'

(send top_level_constant?(:Class, :Module) :new ...) # matches
(send top_level_constant?(:Struct) :new ...) # does not match
```

```ruby
def_node_matcher :top_level_constant?, '(const {nil? (cbase)} %1)'

(send top_level_constant?({:Class :Module}) :new ...) # does not compile
```

Am I missing something?

marcandre:
Right, it is a bit tricky for multiple possibilities, but I think this could work:

```ruby
def_node_matcher :top_level_constant?, '(const {nil? (cbase)} %1)'

def_node_matcher :foo, "(send top_level_constant?(%modules) :new ...)", modules: Set[:Class, :Module]
```

marcandre:
Update: Looks like it should be pretty simple actually to allow any pattern as argument. Implementation is done, I need tests now :smile:, but time for 🛌

tejasbubane:
> Update: Looks like it should be pretty simple actually to allow any pattern as argument. Implementation is done, I need tests now 😄, but time for 🛌

@marcandre Did you implement this?

marcandre:
> > Update: Looks like it should be pretty simple actually to allow any pattern as argument. Implementation is done, I need tests now 😄, but time for 🛌
> 
> @marcandre Did you implement this?

[Yes! 🎉](https://github.com/rubocop-hq/rubocop-ast/pull/64)

lib/rubocop/ast/node.rb
```diff
@@ -510,6 +510,22 @@ def guard_clause?
...
+      def_node_matcher :class_definition?, <<~PATTERN
+        {(class ...)
+         (sclass ...)
+         (casgn _ _ (block (send #global_const?({:Struct :Class}) :new ...) ...))}
+      PATTERN
+
+      def_node_matcher :module_definition?, <<~PATTERN
+        {(module ...)
+         (casgn _ _ (block (send #global_const?(:Module) :new ...) ...))
+         (send nil? :include (block (send #global_const?(:Module) :new ...) ...))}
```
marcandre, line 526: Maybe best to simply check for `(block ...`, no? It could be `var = Module.new {}`, `prepend Module.new`, ...

Same for `class_definition?`

tejasbubane:
@marcandre Changed to capture `body`. The CI failures seem to be related to rubocop-rspec: https://github.com/rubocop-hq/rubocop-rspec/issues/974

marcandre:
All green now 🎉, merging. Great work! 💪

PR #31, opened 2020-06-11
NodePattern: Use `param === node` to match params.

marcandre:
Used to be `node == param`.

For most classes, `===` is the same as `==`, and `==` is symmetric, so this has no impact, except for `Regexp`, `Class`, `Proc`, `Set`, `Range`, none of which can exist in the AST.

This opens many cool patterns though:

* `(hash <(pair (_ %1) $_)>...)` called with a regexp to match a symbol/string key

* `(send _ %1 ...)` called with `Set` of possible method names to match.

etc.

Thanks to @pirj for [bringing this to my attention](https://github.com/rubocop-hq/rubocop-rspec/pull/804)

### DIFF
--- a/docs/modules/ROOT/pages/node_pattern.adoc
+++ b/docs/modules/ROOT/pages/node_pattern.adoc
@@ -374,6 +374,41 @@ def_node_matcher :initializing_with_user?, <<~PATTERN
 PATTERN
 ----
 
+== `%` for arguments
+
+Arguments can be passed to matchers, either as external method arguments,
+or to be used to compare elements. An example of method argument:
+
+[source,ruby]
+----
+def multiple_of?(n, factor)
+  n % factor == 0
+end
+
+def_node_matcher :int_node_multiple?, '(int #multiple_of?(%1))'
+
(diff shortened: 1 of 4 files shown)


docs/modules/ROOT/pages/node_pattern.adoc
```diff
@@ -374,6 +374,41 @@ def_node_matcher :initializing_with_user?, <<~PATTERN
 PATTERN
 ----
 
+== `%` for arguments
```
pirj, line 377: 👍 I was looking for this section today.

pirj:
Works as expected 👍 https://github.com/rubocop-hq/rubocop-rspec/pull/930/files#diff-2aecb8a103588de3553f5448956d701fR87

marcandre:
I'll check about accepting `%MY_CONSTANT` too, should make [things like this](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/lint/useless_comparison.rb#L17-L18) both nicer and more efficient I think.

PR #35, opened 2020-06-14
NodePattern: Add Constants and keyword parameters

marcandre:
This PR allows use of constants in a `NodePattern`, and allows keyword parameters (in addition to the existing sequential parameters).

Example [from the main gem](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/lint/useless_comparison.rb#L15-L18)

```ruby
-       OPS = %w[== === != < > <= >= <=>].freeze
+       OPS = %i[== === != < > <= >= <=>].to_set.freeze

        def_node_matcher :useless_comparison?,
-                        "(send $_match {:#{OPS.join(' :')}} $_match)"
+                        "(send $_match %OPS $_match)"
```

This would also be faster (doing a single hash lookup instead of 8 comparisons, see #29)

The keyword parameters is for cases where positional parameters wouldn't be as clear, but also because I would like to make it possible to specify some keyword parameters at the definition too (upcoming PR)

For example, if one doesn't care to define a constant, one could write:

```ruby
       def_node_matcher :useless_comparison?,
-                         "(send $_match {:#{OPS.join(' :')}} $_match)"
+                         "(send $_match %ops $_match)",
+                         ops: %i[== === != < > <= >= <=>].to_set.freeze
```

This could be overridden when calling the matcher, e.g.
```ruby
useless_comparison?(node, ops: ...)
```

### DIFF
--- a/docs/modules/ROOT/pages/node_pattern.adoc
+++ b/docs/modules/ROOT/pages/node_pattern.adoc
@@ -413,6 +413,29 @@ NOTE: `Array#===` will never match a single node element (so don't pass arrays),
 but `Set#===` is an alias to `Set#include?` (Ruby 2.5+ only), and so can be
 very useful to match within many possible literals / Nodes.
 
+== `%param_name` for named parameters
+
+Arguments can be passed as named parameters. They will be matched using `===`
+(see `%` above).
+
+Contrary to positional arguments, defaults values can be passed to
+`def_node_matcher` and `def_node_search`:
+
+[source,ruby]
+----
+def_node_matcher :interesting_call?, '(send _ %method ...)',
+                 method: Set[:transform_values, :transform_keys,
+                             :transform_values!, :transform_keys!,
+                             :to_h].freeze
+
+# Usage:
+
+interesting_call?(node) # use the default methods
+interesting_call?(node, method: /^transform/) # match anything starting with 'transform'
+----
+
+Named parameters as arguments to custom methods are also supported.
+
 == `nil` or `nil?`
 
 Take a special attention to nil behavior:
--- a/docs/modules/ROOT/pages/node_pattern.adoc
+++ b/docs/modules/ROOT/pages/node_pattern.adoc
@@ -436,6 +436,23 @@ interesting_call?(node, method: /^transform/) # match anything starting with 'tr
 
 Named parameters as arguments to custom methods are also supported.
 
+== `%CONST` for constants
+
+Constants can be included in patterns. They will be matched using `===`, so
++Regexp+ / +Set+ / +Proc+ can be used in addition to literals and +Nodes+:
+
+[source,ruby]
+----
+SOME_CALLS = Set[:transform_values, :transform_keys,
+                 :transform_values!, :transform_keys!,
+                 :to_h].freeze
+
+def_node_matcher :interesting_call?, '(send _ %SOME_CALLS ...)'
+
+----
+
+Constants as arguments to custom methods are also supported.
+
 == `nil` or `nil?`
 
 Take a special attention to nil behavior:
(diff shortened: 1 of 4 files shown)


marcandre:
PS: Still missing doc for `%keyword_parameter` form.

/cc @bbatsov

marcandre:
PR finished, defaults added to `def_node_pattern/search`, documentation added too.

docs/modules/ROOT/pages/node_pattern.adoc
```diff
@@ -409,6 +409,43 @@ has_user_data?(node, /^pass(word)?$/i)
...
 
+== `%keyword` for named parameters
+
+Arguments can be passed as named parameters. They will be matched using `===`.
+Defaults values can be passed to `def_node_matcher` and `def_node_search`.
+
+[source,ruby]
+----
+def_node_matcher :interesting_call?, '(send _ %method ...)',
+                 method: Set[:transform_values, :transform_keys,
```
bbatsov, line 420: Might be a good idea to accept arrays here as well and just implicitly convert them to sets. Seems to me this will result in nicer end user API, if it's more flexible.

docs/modules/ROOT/pages/node_pattern.adoc
```diff
@@ -409,6 +409,43 @@ has_user_data?(node, /^pass(word)?$/i)
 has_user_data?(node, ->(key) { # return true or false depending on key })
 ----
 
+== `%keyword` for named parameters
```
bbatsov, line 412: Not sure if `keyword` is the best term in this context. Maybe something like `named_capture` or `param_name`/`name` would be better? Not sure.

docs/modules/ROOT/pages/node_pattern.adoc
```diff
@@ -409,6 +409,43 @@ has_user_data?(node, /^pass(word)?$/i)
...
+                             :transform_values!, :transform_keys!,
+                             :to_h].freeze
+
+# Usage:
+
+interesting_call?(node) # use the default methods
+interesting_call?(node, method: /^transform/) # match anything starting with 'transform'
+----
+
+Named parameters as arguments to custom methods are not supported.
```
bbatsov, line 430: You can put this in a "NOTE:" so it would stand out.

marcandre: Sure. I didn't support them because I didn't see the need, but now that I think about it, it's trivial to support so probably I should.

docs/modules/ROOT/pages/node_pattern.adoc
marcandre: `%param_name` is good :+1:

docs/modules/ROOT/pages/node_pattern.adoc
marcandre: Aren't you aware of my machiavelic plan to convert the whole world to Sets? ;-)

I like that currently the API is simple, we use `<your object> === <node element>`

It would be easy to convert the values to Sets at the level of `def_node_matcher`, but it would be a bad idea to do it at the matcher call level, so I'm afraid there would be confusion...

I should add a note in the doc that `Array#===` will never work on a AST element.

docs/modules/ROOT/pages/node_pattern.adoc
marcandre: I added support. Implementation took about 2 minutes, +5 for tests +5 for a refactor to make RuboCop happy +10 to open a [bug report](https://github.com/rubocop-hq/rubocop/issues/8158) 😆

marcandre:
Changelog entries added, PR ready as far as I'm concerned.

docs/modules/ROOT/pages/node_pattern.adoc
marcandre: Doc [amended on `master`](https://github.com/rubocop-hq/rubocop-ast/commit/c6db1249cbdf4e0e69ede2825a7f2592b8959a44)

marcandre:
Rebased.

PR #37, opened 2020-06-14
Add `enumerable_method?` for `MethodIdentifierPredicates`

fatkodima:
For context - https://github.com/rubocop-hq/rubocop-performance/pull/140/files#r439732580

This was already needed several times: [one](https://github.com/rubocop-hq/rubocop-performance/pull/140/files#diff-81a815a5887c41f8974ad4d9d34d11f7R41), [two](https://github.com/rubocop-hq/rubocop-performance/pull/127/files#diff-6e1da1392214a67f5b4f8919f93f9b20R31)

### DIFF
--- a/lib/rubocop/ast/node/mixin/method_identifier_predicates.rb
+++ b/lib/rubocop/ast/node/mixin/method_identifier_predicates.rb
@@ -12,6 +12,8 @@ module RuboCop
                               map reduce reject reject! reverse_each select
                               select! times upto].freeze
 
+      ENUMERABLE_METHODS = (Enumerable.instance_methods + [:each]).freeze
+
       # http://phrogz.net/programmingruby/language.html#table_18.4
       OPERATOR_METHODS = %i[| ^ & <=> == === =~ > >= < <= << >> + - * /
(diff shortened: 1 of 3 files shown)


marcandre:
Good work. Another good example for #29 to avoid 61 loops all the time.

lib/rubocop/ast/node/mixin/method_identifier_predicates.rb
```diff
@@ -12,6 +12,8 @@ module MethodIdentifierPredicates
                               map reduce reject reject! reverse_each select
                               select! times upto].freeze
 
+      ENUMERABLE_METHODS = (Enumerable.instance_methods + [:each]).freeze
```
Drenmi, line 15: There's a small chance this might cause an issue due to a mismatch between Ruby version RuboCop is run with versus the version of the code inspected, but I think it's a small enough concern that this is still okay.

marcandre: That's a good point. Might be best to list them explicitly.

PR #38, opened 2020-06-14
Add helpers allowing to check whether the method is a nonmutating operator method or a nonmutating method

fatkodima:
Moved from https://github.com/rubocop-hq/rubocop/pull/8155

For context https://github.com/rubocop-hq/rubocop-performance/pull/140/files#r439732580

Bullet number 4

### DIFF
--- a/lib/rubocop/ast/node/mixin/method_identifier_predicates.rb
+++ b/lib/rubocop/ast/node/mixin/method_identifier_predicates.rb
@@ -6,17 +6,62 @@ module RuboCop
     # `send`, `csend`, `def`, `defs`, `super`, `zsuper`
     #
     # @note this mixin expects `#method_name` and `#receiver` to be implemented
-    module MethodIdentifierPredicates
+    module MethodIdentifierPredicates # rubocop:disable Metrics/ModuleLength
       ENUMERATOR_METHODS = %i[collect collect_concat detect downto each
                               find find_all find_index inject loop map!
                               map reduce reject reject! reverse_each select
-                              select! times upto].freeze
+                              select! times upto].to_set.freeze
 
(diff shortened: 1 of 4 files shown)


lib/rubocop/ast/node/mixin/method_identifier_predicates.rb
```diff
@@ -1,22 +1,69 @@
 # frozen_string_literal: true
 
+require 'set'
```
marcandre, line 3: The gem `parser` (and maybe others) already `require 'set'`. I'd remove it entirely, but if you prefer having it for clarity, it belongs in `lib/reubocop/ast.rb`.

spec/rubocop/ast/send_node_spec.rb
```diff
@@ -736,6 +736,126 @@ module Foo
     end
   end
 
+  describe '#nonmutating_binary_operator_method?' do
+    context 'with a nonmutating binary operator method' do
+      let(:source) { 'foo + bar' }
+
+      it { expect(send_node.nonmutating_binary_operator_method?).to be_truthy }
```
marcandre, line 743: Please use `be(true)` and `be(false)` until we can do better

lib/rubocop/ast/node/mixin/method_identifier_predicates.rb
fatkodima: > but if you prefer having it for clarity

yes, this. 
And imo it is better to not depend on something else required it.
Moved to suggested location.

spec/rubocop/ast/send_node_spec.rb
fatkodima: Yes, the whole test codebase is not written the way I would prefer, but I was consistent with what we have 😄

PR #40, opened 2020-06-15
Fix source and lines handling for heredocs

fatkodima:
[Parser's heredoc format]( https://github.com/whitequark/parser/blob/master/doc/AST_FORMAT.md#here-document):

```
(str "foo\nbar\n")
'<<HERE␊foo␊bar␊HERE'
 ~~~~~~ expression
        ~~~~~~~~ heredoc_body
                ~~~~ heredoc_end
```

Right now, heredocs are not handled correctly, imo. Every updated method in this PR in the original implementation only deals with `expression` part. And so returns incorrect values (look at tests for expected behavior).

Rubocop needs a couple of changes (2-3 very small) due to new behavior to let the CI pass. But this gem depends on `rubocop`, and `rubocop` depends on this gem, so I need and advise on how to proceed and in what order.


marcandre:
Heredocs are complicated. The notion of 'source' of a heredoc is strange as the source is disconnected.

It's important to realize how disconnected it is...

```ruby
foo(<<~RUBY, other, arguments(:here), not(related_to_here_doc))
  inside heredoc
RUBY
```

I don't think that including `, other, arguments(:here), not(related_to_here_doc))`in the `source`  will be of much help to deal with heredocs...

fatkodima:
Yeah, agreed.
But what about `#first_line`, `#last_line` and (not present in this pr) `#line_count` methods?
I'm working on a pr into main repository, where I need to get the number of lines in heredoc, so specially interested in `#line_count` method.

fatkodima:
Feel free to ping me if there would be any interest in this.

marcandre:
Thanks for the ping, sorry I didn't answer before.

I'm 👍 on `#first_line, `#last_line`, and (not present in this pr) `#line_count`, and anything else that can be useful to you or others and is well defined :-)
Please have tests for complex cases (only), e.g.
```
x = foo(<<~FIRST, <<~SECOND)
  part of first heredoc
FIRST
  this is
  part of second heredoc
SECOND
```

marcandre:
We can consider `line_range` (instead?), and maybe they should be named `body_line_range` or `heredoc_line_range` to underline that it is the range of the text itself, not of the begin/end markers

fatkodima:
On master branch, for heredocs:
`StrNode#first_line` works as expected, 
fixed `#last_line` to return the correct line (of the ending marker) - on master always returns the line of first marker
fixed `#line_count` to return the correct line count (2 markers + of content) - on master always returns 1 as lines in first marker

PR #45, opened 2020-06-19
Run internal investigation using master only. Fix disabling

marcandre:
There are cases where it won't be possible to pass internal investigation for both `master` and official rubocop (like now that CyclomaticComplexity has been tweaked), so run internal investigation only on `master`.


marcandre:
Wow, github action's syntax is weak! Won't support `a && b && c` 🤷‍♂️

PR #49, opened 2020-06-25
Support "Endless method definition" syntax for Ruby 2.8 (3.0)

koic:
This PR supports "Endless method definition" syntax for Ruby 2.8 (3.0).
Parser gem supports this syntax by https://github.com/whitequark/parser/pull/676.

Ref: https://bugs.ruby-lang.org/issues/16746

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#49](https://github.com/rubocop-hq/rubocop-ast/pull/49): Support "Endless method definition" syntax for Ruby 2.8 (3.0). ([@koic][])
+
 ## 0.1.0 (2020-06-26)
 
 ### New features
--- a/lib/rubocop/ast/builder.rb
+++ b/lib/rubocop/ast/builder.rb
@@ -26,6 +26,7 @@ module RuboCop
         case:         CaseNode,
         class:        ClassNode,
         def:          DefNode,
+        def_e:        DefNode,
         defined?:     DefinedNode,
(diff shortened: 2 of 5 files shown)


bbatsov:
I think we should also add an `endless?` predicate to the node extension.

marcandre:
I commented in whitequark/parser#676 that I hope that we can get a simple `def` (and also in https://bugs.ruby-lang.org/issues/16746 that I hope this is reverted...)
I think we should wait for this feature to be confirmed, or at at least for `parser`'s response?

koic:
Thanks for the discussion at Parser. I thought that it could be introduced to RubyCop AST experimentally, referring to the fact that Ruby Next supports it.
https://github.com/ruby-next/ruby-next

> I think we should wait for this feature to be confirmed, or at at least for parser's response?

At least I think we can wait for Parser's response.

> https://bugs.ruby-lang.org/issues/16746 that I hope this is reverted...

To be honest, I'm also wondering about adding this syntax to Ruby 😅 

Apart from that I added `endless?` predicate. Thank you.

marcandre:
So `parser` will output `def` after all. We could add an `endless?` predicate that would check the presence of `loc.end`

koic:
Yeah, I will update this PR when Parser 2.7.1.5 (or Parser 2.7.2.0) is released. Thank you for your kindness advice!

koic:
This PR will be closed due to prefer #129. Thank you @marcandre!

marcandre:
> This PR will be closed due to prefer #129. Thank you @marcandre!

My pleasure 😄

PR #52, opened 2020-07-01
[Fixes #44] Use parser's new emit_forward_arg by default

### LINKED ISSUE #44: Parser and emit_forward_arg
I'd like to use the latest version of parser with `emit_forward_arg` turned on.

We could avoid [this "hack" of redefining `to_a` as different from `children`](https://github.com/rubocop-hq/rubocop-ast/blob/master/lib/rubocop/ast/node/forward_args_node.rb#L13-L15)

```sh
$ ruby-parse --legacy-forward-arg -e "def foo(...);end; def bar(*rest); end"
(begin
  (def :foo
    (forward-args) nil)
  (def :bar
    (args
      (restarg :rest)) nil))

$ ruby-parse -e "def foo(...);end; def bar(*rest); end"
(begin
  (def :foo
    (args
      (forward-arg)) nil)
  (def :bar
    (args
      (restarg :rest)) nil))
```

The later form is much improved as an argument signature is always inside an `args` node.

I doubt it would be much incompatibility, but v1 would be perfect for this.

Comments / objections?

marcandre:


lib/rubocop/ast/builder.rb
```diff
@@ -14,6 +14,8 @@ module AST
     #   parser = Parser::Ruby25.new(builder)
     #   root_node = parser.parse(buffer)
     class Builder < Parser::Builders::Default
+      self.emit_forward_arg = true
```
bbatsov, line 17: Shouldn't we leave it to RuboCop to set this, for the sake of consistency in `rubocop-ast`?

README.md
```diff
@@ -35,8 +35,9 @@ See the [docs site](https://docs.rubocop.org/rubocop-ast) for more details.
 
 ### Parser compatibility switches
 
-The main `RuboCop` gem uses [legacy AST output from parser](https://github.com/whitequark/parser/#usage).
-This gem is meant to be compatible with all settings. For example, to have `-> { ... }` emitted
+This gem, by default, uses most [legacy AST output from parser](https://github.com/whitequark/parser/#usage), except for `emit_forward_arg` which is set to `true`.
+
+The main `RuboCop` gem uses these defaults (and is currently only compatible with these), but this gem can be used separately from `RuboCop` and is meant to be compatible with all settings. For example, to have `-> { ... }` emitted
```
bbatsov, line 40: I think a more salient point here is that most tools using `parser` use the legacy format, that's why it's a safer default.

lib/rubocop/ast/builder.rb
marcandre: Contrary to the other `emit_`, this is really fixing an error in the AST. We are < 1.0 so I think it's best to set this default now.

README.md
marcandre: > I think a more salient point here is that most tools using parser use the legacy format

Quite possible (by choice or because they aren't aware of it). I know some are using the "modern" AST  (`unparser`, `ruby-next`), but my tool (`DeepCover`) doesn't, by ignorance first and by choice now 😆 

> that's why it's a safer default.

I'm not sure what you mean by "safer".

My objective is to be useful. Have you actually looked closely at the `emit_...` flags?

In my opinion, while they reflect an actual difference in the language, all of them are insignificant over-complexifications for the vast majority of tools and Rubyists.

For example: `-> {}` and `lambda {}` are indeed different in theory because one could, for example, redefine the method `lambda` while the first form woudn't be affected. I'd venture to say that the vast majority of `parser` uses don't care one bit about that difference. For `RuboCop` there might be a single cop caring about it (fixing the preference of one form over the other); all other cops probably prefer the simplicity of having to deal with the same AST for both constructs.

Some are [even more obscure](https://github.com/whitequark/parser/issues/341). I personally much prefer an AST of `(send _ :[]= ...)`; it's how I think about `foo[42]= :bar`. I've yet to see `foo.[]=(42, :bar)` in the wild and it's quite sad that it forced `parser` to introduce a different AST for this.

`emit_forward` is different. It was actually [a bad decision](https://github.com/whitequark/parser/pull/710#issuecomment-642728715) and the AST emitted without it is of a strange structure (`forward-args` not inside an `args`, ...) and is not forward compatible with 2.8. My opinion is that it should be turned on by default by all tools that are aware of it.

Are there `emit_...` that you feel are actually useful when turned on?

lib/rubocop/ast/builder.rb
bbatsov: Fair enough.

README.md
bbatsov: > that's why it's a safer default.

I meant it's less likely to surprise someone with the modern parser output. Maybe that's not a big problem in practice. 

> My objective is to be useful. Have you actually looked closely at the emit_... flags?

Yeah, I have. Generally most of the changes didn't add much value for us, that's why I never felt compelled to update the code. The biggest practical issue is obviously the big number of extensions that would suffer if we update the format, so for such changes we should always have a good reason. I'm fine with updating `forward-args`, as it has a relatively small impact, but I want to make sure people understand our rationale.

PR #53, opened 2020-07-04
Simplify and optimize ArrayNode#values

marcandre:


Drenmi:
This is technically a breaking change, because `#children` also returns non-node members, whereas `#each_child_node` only returns nodes.

marcandre:
> This is technically a breaking change, because `#children` also returns non-node members, whereas `#each_child_node` only returns nodes.

In an array literal? Can you provide an example where there would be a non-node member?

PR #55, opened 2020-07-04
Fix `ProcessedSource#commented?` for multi-line ranges. Add `ProcessedSource#commented_line?`

marcandre:
`ProcessedSource#commented?` was only considering the beginning line of the given range

### DIFF
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -99,9 +99,19 @@ module RuboCop
         ast.nil?
       end
 
-      def commented?(source_range)
-        comment_lines.include?(source_range.line)
+      # @return [Boolean] if the given line number has a comment.
+      def line_with_comment?(line)
+        comment_lines.include?(line)
       end
+
+      # @return [Boolean] if any of the lines in the given `source_range` has a comment.
+      def contains_comment?(source_range)
+        (source_range.line..source_range.last_line).any? do |line|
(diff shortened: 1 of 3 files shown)


lib/rubocop/ast/processed_source.rb
```diff
@@ -95,8 +95,16 @@ def blank?
         ast.nil?
       end
 
+      # @return [Boolean] if the given line number has a comment.
+      def commented_line?(line)
```
bbatsov, line 99: I think the name is a bit misleading (it sounds like a line that was commented out), but it's a good change. I guess you just tried to mimic the `commented?` name we've employed before. Perhaps it should have been something like `contains_comments?` or whatever.

marcandre: I agree, I should have thought of a better name, especially that `util` in `rubocop` has `comment_line?` which is true only for full lines.
So which is best: `contains_comment?`, or `line_with_comment?`.

bbatsov: You can make a method `contains_comment?` that acts differently on line numbers and source ranges. :wink: If not - I'd go with `line_with_comment?`, as I suggested `contains_comment?` as an alternative name for `commented?`.

marcandre: Updated with `line_with_comment?` and `contains_comment?` (with deprecated alias `commented?`)

PR #63, opened 2020-07-11
NodePattern: Allow full expressions as function arguments

marcandre:


docs/modules/ROOT/pages/node_pattern.adoc
```diff
@@ -316,6 +316,30 @@ We can use the `#prime?` method directly in the expression:
 (int #prime?)
 ----
 
+== Arguments for predicate and external methods
```
bbatsov, line 319: predicates and external methods? I'm confused by this terminology. :-)

bbatsov:
Looks cool to me. I only find the terminology confusing - e.g. what's special about predicate methods.

docs/modules/ROOT/pages/node_pattern.adoc
marcandre: Right, I'll replace "external methods" with "funcall", which is what is used elsewhere (`predicate?`, `#funcall`)

PR #65, opened 2020-07-13
Simplify and optimize.

marcandre:
This reduces memory allocation from 19MB to 1.9MB (when run on cop/mixin).

Main issue remains that it is called way too many times though


bbatsov:
> This reduces memory allocation from 19MB to 1.9MB (when run on cop/mixin).

Small change, huge impact!

marcandre:
My disappointment is that there are 3 separate anti-patterns in this code... I created the issues so we try to detect them.

PR #69, opened 2020-07-18
Remove `RetryNode`.

marcandre:
This node was severely broken, e.g. `#arguments?` raised a `NoMethodError`.
I find it has nothing to do with a `MethodDispatchMethod` (since it
doesn't call a method), nor a `ParameterizedNode` (since it accepts no
argument); best remove the node altogether.


marcandre:
See #71 for discussion if it would be useful to have `receiver` return `nil` and `method_name` return `:retry`

PR #70, opened 2020-07-18
BreakNode: Fix arguments

marcandre:


marcandre:
Alternative would be to have `#receiver` return `nil` and `#method_name` return `:return / :break / :next`, like we do for `yield`, `super` & `defined?`.

My first impression is that while `yield`, `super` and `defined?` are keywords, they act more like method call. `return/break/next` change the control flow so I don't feel that they should be considered similarly (same as `retry`), but those with more experience writing cops can chime in.

PR #75, opened 2020-07-22
Fix `ProcessedSource` for cases with syntax errors

marcandre:
In some circumstances, `ast` could be `false` instead of `nil`, `comments` and `tokens` could be `nil` instead of `[]`.

### DIFF
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -168,10 +168,12 @@ module RuboCop
           ast ||= nil # force `false` to `nil`, see https://github.com/whitequark/parser/pull/722
         rescue Parser::SyntaxError
           # All errors are in diagnostics. No need to handle exception.
+          comments = []
+          tokens = []
         end
 
         ast&.complete!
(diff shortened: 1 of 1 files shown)


lib/rubocop/ast/processed_source.rb
```diff
@@ -165,13 +165,15 @@ def parse(source, ruby_version)
...
+          ast ||= nil # force `false` to `nil`, see https://github.com/whitequark/parser/pull/722
         rescue Parser::SyntaxError
           # All errors are in diagnostics. No need to handle exception.
+          comments = []
+          tokens = []
         end
 
-        tokens = tokens.map { |t| Token.from_parser_token(t) } if tokens
+        ast&.complete!
+        tokens = tokens.map { |t| Token.from_parser_token(t) }
```
bquorning, line 176: Is `tokens` an object shared with other code? Or could `tokens.map!` be a viable optimization here?

marcandre: Sure, I'll change it to `map!`

marcandre: Done 👍

PR #83, opened 2020-07-26
Review comments API

marcandre:
This PR:
* Adds `ProcessedSource#comment_at_line`
* Adds `ProcessedSource#each_comment_in_lines`

It deprecates `ProcessedSource#comments_before_line` (which is imo a big code smell).

I can't believe how much time I spent to arrive to this quite small PR; it's like version 5.0, I was really slow to get to this API.

I have reviewed / fixed all uses of any comment-accessing method in the main gem. The PR is upcomming. My main take is that any PR that autocorrects any range that could span more than one line must have specs with comments *on each single line* (close to that range).

### DIFF
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -76,7 +76,7 @@ module RuboCop
         comments.each(&block)
       end
 
-      # @deprecated Use `comments.find` or `comment_at_line`
+      # @deprecated Use `comment_at_line`, `each_comment_in_lines`, or `comments.find`
       def find_comment(&block)
         comments.find(&block)
       end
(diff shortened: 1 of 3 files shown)


marcandre:
@bbatsov, do you prefer I wait for your approval for this kind of PR?

bbatsov:
Sorry about the slow response. I was on vacation for a week and I'm behind on a ton of things. The changes seem fine to me. 

> @bbatsov, do you prefer I wait for your approval for this kind of PR?

If you're very confident about some changes you don't really need my approval. I don't want to be a bottleneck for progress. :-)

lib/rubocop/ast/processed_source.rb
```diff
@@ -99,22 +99,39 @@ def blank?
         ast.nil?
       end
 
+      # @return [Comment, nil] the comment at that line, if any.
+      def comment_at_line(line)
```
bbatsov, line 103: I guess this covers both whole line comments and inline comments, right?

marcandre: Yes! API that applies only to whole line comments should be pretty rare imo

marcandre:
>  I was on vacation for a week

Great, hope you had a good time!

marcandre:
> If you're very confident about some changes you don't really need my approval.

Only issue with this is I'm a very confident person 😅, and I don't always put enough emphasis on naming / doc.

Given these, I'll merge #82 which I consider obvious, and I'll wait for #72 which is on hold because of the name of a module meant for internal use (your input appreciated)...

PR #84, opened 2020-07-26
Add `Parser::Source::Range#line_span`

marcandre:
I've seen about 20 occurrences of `range.{first}line..{.}range.last_line` or calls to `line_range(node)` (defined in `util.rb`)

I feel the name `line_span` is more descriptive and is less confusing with `Source::Buffer#line_range` which returns a range of columns for the given line.

I notice that there currently are no extension to `parser`'s classes, but this one seems reasonable.

The result can be particularly useful when dealing with comments, say with `each_comment_in_lines` of #83


bbatsov:
> I notice that there currently are no extension to parser's classes, but this one seems reasonable.

Agreed.

PR #85, opened 2020-08-01
Add `IntNode#value` and `FloatNode#value`

fatkodima:


marcandre:
Definitely!

Thanks 👍

PR #87, opened 2020-08-01
Add `CaseNode#branches`

marcandre:
I was surprised that `IfNode#branches` returns a array with a single item for `if foo; else; end`. I was expecting `[nil, nil]`, not just `[nil]`.

I mimicked this behavior for `CaseNode` as it seems natural to have this symmetry, but I wonder which is best. Anyone has an idea? cc/ @fatkodima

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [#83](https://github.com/rubocop-hq/rubocop-ast/pull/83): Add `ProcessedSource#comment_at_line` ([@marcandre][])
 * [#83](https://github.com/rubocop-hq/rubocop-ast/pull/83): Add `ProcessedSource#each_comment_in_lines` ([@marcandre][])
 * [#84](https://github.com/rubocop-hq/rubocop-ast/pull/84): Add `Source::Range#line_span` ([@marcandre][])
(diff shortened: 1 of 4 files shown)


fatkodima:
Symmetry is good, but it seems to me also that it should return `[nil, nil]` instead of just `[nil]` for provided example.

marcandre:
There are different possibilities for `IfNode#branches`

1) always return the last `else` branch, even if it is only implicit:
```ruby
if foo; end # => [nil, nil]
```
2) return the last `else` branch only if there's an explicit keyword
```ruby
if foo; end # => [nil]
if foo; else; end # => [nil, nil]
```
3) return the last `else` branch only if there keyword *and* some code in there [current]
```ruby
if foo; else; end # => [nil]
if foo; else; 42; end # => [nil, s(:int, 42)]
```

I like #3 the least.

The same question exist for `case` and I'd like the answer to be the same... @fatkodima you prefer #2 or #1

fatkodima:
No. 2 seems the most logical to me.

marcandre:
Thank you :-)

I made the tweak, and not a single spec fails, in `rubocop-ast` or `rubocop`... I put this as "potentially breaking" in the Changelog.

PR #92, opened 2020-08-05
Add `ProcessedSource#tokens_within`, `ProcessedSource#first_token_of` and `ProcessedSource#last_token_of`

fatkodima:
Context - https://github.com/rubocop-hq/rubocop/pull/8450

That was merged a bit fast 😄  - will send PR with refactorings when this will be released.

### DIFF
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -159,6 +159,20 @@ module RuboCop
           .length
       end
 
+      def tokens_within(range_or_node)
+        begin_index = first_token_index(range_or_node)
+        end_index = last_token_index(range_or_node)
+        sorted_tokens[begin_index..end_index]
+      end
+
+      def first_token_of(range_or_node)
+        sorted_tokens[first_token_index(range_or_node)]
+      end
+
+      def last_token_of(range_or_node)
+        sorted_tokens[last_token_index(range_or_node)]
+      end
+
       private
 
       def comment_index
@@ -238,6 +252,31 @@ module RuboCop
           parser.diagnostics.consumer = lambda do |diagnostic|
             @diagnostics << diagnostic
           end
+        end
+      end
+
+      def first_token_index(range_or_node)
+        begin_pos = source_range(range_or_node).begin_pos
+        sorted_tokens.bsearch_index { |token| token.begin_pos >= begin_pos }
+      end
+
+      def last_token_index(range_or_node)
+        end_pos = source_range(range_or_node).end_pos
+        sorted_tokens.bsearch_index { |token| token.end_pos >= end_pos }
+      end
+
+      # The tokens list is always sorted by token position, except for cases when heredoc
+      # is passed as a method argument. In this case tokens are interleaved by
+      # heredoc contents' tokens.
+      def sorted_tokens
+        @sorted_tokens ||= tokens.sort_by(&:begin_pos)
+      end
+
+      def source_range(range_or_node)
+        if range_or_node.respond_to?(:source_range)
+          range_or_node.source_range
+        else
+          range_or_node
         end
       end
     end
--- a/spec/rubocop/ast/processed_source_spec.rb
+++ b/spec/rubocop/ast/processed_source_spec.rb
@@ -10,6 +10,7 @@ RSpec.describe RuboCop::AST::ProcessedSource do
     end
     some_method
   RUBY
+  let(:ast) { processed_source.ast }
   let(:path) { 'ast/and_node_spec.rb' }
 
   shared_context 'invalid encoding source' do
@@ -292,7 +293,6 @@ RSpec.describe RuboCop::AST::ProcessedSource do
     describe '#contains_comment?' do
       subject(:commented) { processed_source.contains_comment?(range) }
 
-      let(:ast) { processed_source.ast }
       let(:array) { ast }
       let(:hash) { array.children[1] }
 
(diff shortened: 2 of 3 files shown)


lib/rubocop/ast/processed_source.rb
```diff
@@ -159,6 +160,37 @@ def line_indentation(line_number)
           .length
       end
 
+      def node_tokens(node)
+        @node_tokens ||= {}.compare_by_identity
```
marcandre, line 164: I think a worthwhile method would be `def tokens_within(source_range)`.

Was this cache actually helpful in your tests, in the end? If so, then we might not even need `node_tokens`, otherwise we could have both.

lib/rubocop/ast/processed_source.rb
```diff
@@ -159,6 +160,37 @@ def line_indentation(line_number)
...
+        return @node_tokens[node] if @node_tokens[node]
+
+        begin_index = node_first_token_index(node)
+        end_index   = node_last_token_index(node)
+        @node_tokens[node] = tokens[begin_index..end_index]
+      end
+
+      def node_first_token(node)
+        node_tokens(node).first
```
marcandre, line 173: Are we sure we need the method at all? Implementation is not efficient...

lib/rubocop/ast/processed_source.rb
```diff
@@ -159,6 +160,37 @@ def line_indentation(line_number)
...
+      end
+
+      def node_last_token(node)
+        node_tokens(node).last
```
marcandre, line 177: Are we sure we need the method at all? Implementation is not efficient...

lib/rubocop/ast/processed_source.rb
```diff
@@ -240,6 +272,13 @@ def create_parser(ruby_version)
           end
         end
       end
+
+      def sort_tokens_if_needed
+        return if @tokens_sorted
+
+        tokens.sort_by!(&:begin_pos)
```
marcandre, line 279: Tokens should be frozen. I think it's not the best idea to mutate it anyways; best to have a method `sorted_tokens` I think. FYI, it doesn't require more memory allocation, only both arrays won't be garbage collectable while we process the file instead of one.

spec/rubocop/ast/processed_source_spec.rb
```diff
@@ -464,4 +464,98 @@ def some_method
...
+
+      it 'returns tokens for node before heredoc' do
+        node = ast.children[0].arguments[1]
+        tokens = processed_source.node_tokens(node)
+
+        expect(tokens.map(&:text)).to eq(['[', 'before', ']'])
+      end
+
+      it 'returns tokens for heredoc node' do
+        node = ast.children[0].arguments[2]
```
marcandre, line 499: Aren't the heredoc's tokens a bigger issue for `#{}` within the heredoc?

lib/rubocop/ast/processed_source.rb
```diff
@@ -159,6 +160,37 @@ def line_indentation(line_number)
...
+
+      def node_first_token(node)
+        node_tokens(node).first
+      end
+
+      def node_last_token(node)
+        node_tokens(node).last
+      end
+
+      def node_first_token_index(node)
```
marcandre, line 180: Will you be using this directly? I'd put it private, or else I would change the api to take a `begin_pos`

lib/rubocop/ast/processed_source.rb
```diff
@@ -159,6 +160,37 @@ def line_indentation(line_number)
...
+        sort_tokens_if_needed
+
+        begin_pos = node.source_range.begin_pos
+        tokens.bsearch_index { |token| token.begin_pos >= begin_pos }
+      end
+
+      def node_last_token_index(node)
+        sort_tokens_if_needed
```
marcandre, line 188: Same.

lib/rubocop/ast/processed_source.rb
fatkodima: Retested with and without cache - it is useless so removed that and `node_tokens` method; added `tokens_within`

lib/rubocop/ast/processed_source.rb
fatkodima: Changed the name and implementation; yes, we need it in several places, like https://github.com/rubocop-hq/rubocop/blob/30655006edf2921c323c9f7ec3883629dbde5179/lib/rubocop/cop/layout/space_inside_string_interpolation.rb#L58-L60 and https://github.com/rubocop-hq/rubocop/blob/30655006edf2921c323c9f7ec3883629dbde5179/lib/rubocop/cop/layout/space_inside_hash_literal_braces.rb#L97-L102

lib/rubocop/ast/processed_source.rb
fatkodima: Same as above

spec/rubocop/ast/processed_source_spec.rb
fatkodima: Didn't get the point. Can you reword/explain more?

lib/rubocop/ast/processed_source.rb
fatkodima: Made private.

spec/rubocop/ast/processed_source_spec.rb
marcandre: I was wondering: would your test fail if the tokens were not sorted? I thought it was for cases where there's a heredoc with for `#{}` that doing the `bsearch` on unsorted tokens might fail.

lib/rubocop/ast/processed_source.rb
```diff
@@ -240,6 +254,21 @@ def create_parser(ruby_version)
           end
         end
       end
+
+      def first_token_index(source_range)
+        sorted_tokens.bsearch_index { |token| token.begin_pos >= source_range.begin_pos }
+      end
+
+      def last_token_index(source_range)
+        sorted_tokens.bsearch_index { |token| token.end_pos >= source_range.end_pos }
```
marcandre, line 263: How about `begin_pos = source_range.begin_pos` before the `bsearch_index`, as a "free" optimization, or passing that instead of `source_range`?

lib/rubocop/ast/processed_source.rb
```diff
@@ -159,6 +159,20 @@ def line_indentation(line_number)
           .length
       end
 
+      def tokens_within(source_range)
```
marcandre, line 162: Last thing: maybe we should accept a `source_range` or a `node`?
I try to avoid these for fast methods, but here a single `if` won't be noticeable. WDYT?

lib/rubocop/ast/processed_source.rb
```diff
@@ -159,6 +159,20 @@ def line_indentation(line_number)
...
+        begin_index = first_token_index(source_range)
+        end_index = last_token_index(source_range)
+        sorted_tokens[begin_index..end_index]
+      end
+
+      def first_token_of(source_range)
```
marcandre, line 168: We could accept `node` or `source_range` here too and in `last_token_of`

spec/rubocop/ast/processed_source_spec.rb
fatkodima: When tokens are not sorted, for some cases, depending on the source, it will fail, sometimes not. It depends on how lucky bsearch will be to find correct locations.
 When they are sorted, it will always succeed.

lib/rubocop/ast/processed_source.rb
fatkodima: Sure, I can extract it into local variable, but it seems that won't give much a difference. For linear search this probably will have noticeable effect, but for bsearch - not sure.

I think, most of the time this method will be used with `Node`s as an argument, so 3 variants of arguments looks equal to me. `source_range` looks more clear to me, but I have no strong opinion on this.

marcandre: I feel supporting `source_range` is important as it is the most generic. There might not be a `Node` with the `source_range` someone needs.

> most of the time this method will be used with Nodes

So let's accept a `source_range`, or anything that `responds_to?(:source_range)` then.

fatkodima:
Updated to accept `source_range` and `Node`.

lib/rubocop/ast/processed_source.rb
```diff
@@ -240,6 +254,31 @@ def create_parser(ruby_version)
...
+
+      # The tokens list is always sorted by token position, except for cases when heredoc
+      # is passed as a method argument. In this case tokens are interleaved by
+      # heredoc contents' tokens.
+      def sorted_tokens
+        @sorted_tokens ||= tokens.sort_by(&:begin_pos)
+      end
+
+      def source_range(range_or_node)
+        if range_or_node.is_a?(RuboCop::AST::Node)
```
marcandre, line 276: I thought that `Comment` responded to `:source_range` but it doesn't. I'll open an issue with `parser` coz it should I think :-)
Was there a reason to prefer checking for a class instead of duck typing with `respond_to? :source_range` as I had proposed?

fatkodima: ok, made it a duck 😄

marcandre:
Perfect. I <3 it. We're ready to merge, but ChangeLog entry must be moved. These conflicts in the Changelog are quite annoying. I'd list the 3 methods added in it btw.

fatkodima:
> These conflicts in the Changelog are quite annoying

Looks like a startup idea 😄 Something like `dependa_bot`, but for changelog entries 😄

marcandre:
Super, thanks! 🎉

PR #93, opened 2020-08-10
Add Node#left_sibling, right_sibling, left_siblings and right_siblings

marcandre:
[Requested](https://github.com/rubocop-hq/rubocop/pull/8474#discussion_r466959357) by @fatkodima


fatkodima:
👍 
I think, just `siblings` may also be useful too.

marcandre:
> 👍
> I think, just `siblings` may also be useful too.

as in `left_siblings + right_siblings`?

fatkodima:
Haven't find any, but maybe, mayyybeee (😄 ) it will be needed in the future.

marcandre:
So I merged this, if we see a lot of `left_siblings + right_siblings`, we'll add `siblings`...

PR #109, opened 2020-09-11
NodePattern: debugging tools

marcandre:
Builds on #105 to add `WithMeta` (i.e. locations and comments) and `Debug` variants, with some debugging rake tasks.

PR #110, opened 2020-09-11
[Fixes #79] NodePattern: Add support for multiple terms unions

### LINKED ISSUE #79: Incomplete tuple matching
This is more of a question than a feature request.
In [`rubocop-rspec`](https://github.com/rubocop-hq/rubocop-rspec)'s cops there are two patterns have redundancy.

1. [`RSpec/SubjectStub`](https://github.com/rubocop-hq/rubocop-rspec/blob/6266a6a2abefc275558f7360148db32ef36e7443/lib/rubocop/cop/rspec/subject_stub.rb#L45)
```
{
  (block (send nil? :subject (sym $_)) ...)
  (block (send nil? $:subject) ...)
}
```

It is supposed to capture `:subject` or the subject's name from:
```ruby
subject { ... }
# or
subject(:name) { ... }
```

What makes me think it's redundant?
 - an equal number of captures
 - same wrapping, e.g. base of the tree is the same

What makes it impossible to turn this into:
```
(block
  (send nil?
    {
      :subject (sym $_)
      $:subject
    }
  )
  ...
)
```
is that `{}` would consider all three expressions as matching options, and there is no way (that I'm aware of) to match two or more subsequent nodes that do not compose a complete sequence (terminology may be distorted, hope the idea is clear though).

2. The same applies to another cop, `RSpec/DescribeClass`:
```
          (block
            [  (send #{RSPEC} :describe $[!const !#string_constant?] ...)
              !(send #{RSPEC} :describe ... (hash <#rails_metadata? ...>)) ]
            ...
          )
```
It can't be:
```
(block
  (send #{RSPEC} :describe
    [
      $[!const !#string_constant?] ...
      ... !(hash <#rails_metadata? ...>)
    ]
  )
  ...
)
```
because there's no way to combine those (newline-separated) elements into tuples.

I would really prefer to keep Node Pattern syntax simple. And, unlike the previously introduced unordered sequence (`<>`) there's much of a compelling benefit since I could only find two cops that would just slightly (and arguably) benefit from that.

On the other hand, `<>` is present already, but it disregards the order. Maybe it's possible somehow to respect the order given some specifier?
Not a real proposal, just thinking aloud:
```
(block (send nil? { <- :subject (sym $_)> $:subject } ) ... )
```
(`Couldn't compile due to invalid token "<".`)

```
(block
  (send #{RSPEC} :describe
     <- $[!const !#string_constant?] ...>
    !<- ... (hash <#rails_metadata? ...>)>
  ) ...)
```

where `<- >` is an "ordered sequence".

I didn't look much into RuboCop core's cops, can't estimate if any of them could benefit from such syntax.

marcandre:
This builds on  #109 and adds `{ multiple term | another multiple term branch }`


marcandre:
Merged 🎉 . To be released tomorrow.

PR #111, opened 2020-09-11
NodePattern: Set optimization

marcandre:
This builds on #110 to optimize `NodePattern` with unions of literals like `(send nil? {:method :other_method})`

```sh
> rake compile '{1 2 3}'
# before:
(1 === node || 2 === node || 3 === node)
# after:
::RuboCop::AST::NodePattern::Sets::SET_1_2_3 === node
```

And `SET_1_2_3` is set to `Set[1, 2, 3]`


marcandre:
Merged 🎉 . To be released tomorrow.

PR #115, opened 2020-09-16
Fix `ConstNode#absolute?` raising `NoMethodError` when the constant is not namespaced

dvandersluis:
Came up in https://github.com/rubocop-hq/rubocop/pull/8738.

When there is no namespace, `each_path.first` is nil, which caused

```
NoMethodError: undefined method `cbase_type?' for nil:NilClass
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+### Bug fixes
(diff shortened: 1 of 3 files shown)


marcandre:
Oops, good catch, thanks!

marcandre:
Released in v0.4.1

PR #120, opened 2020-09-24
Fix macro?

marcandre:
This PR fixes rubocop-hq/rubocop#8765

There were many issues with how `macro_scope?`:
* recursion was badly written, so many cases were not handled well
* `if` did not distinguish if in the conditions part or on of the bodies

It also introduces `Node#parent?`

Builds on #119 

I'll do a final review when I'm less sleepy 😅


marcandre:
Blocked by rubocop-hq/rubocop#8789 and #123

PR #121, opened 2020-09-24
Update from `Parser::Ruby28` to `Parser::Ruby30` for Ruby 3.0 parser

koic:
Parser 2.7.1.5 includes https://github.com/whitequark/parser/pull/729 and this PR updates from `Parser::Ruby28` to `Parser::Ruby30` for Ruby 3.0 parser.

### DIFF
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -229,9 +229,9 @@ module RuboCop
         when 2.7
           require 'parser/ruby27'
           Parser::Ruby27
-        when 2.8
-          require 'parser/ruby28'
-          Parser::Ruby28
+        when 2.8, 3.0
+          require 'parser/ruby30'
+          Parser::Ruby30
         else
           raise ArgumentError,
                 "RuboCop found unknown Ruby version: #{ruby_version.inspect}"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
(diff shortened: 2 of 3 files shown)


koic:
This PR should be synchronized with https://github.com/rubocop-hq/rubocop/pull/8785 for the successful build.

koic:
@bbatsov @marcandre Can you please release RuboCop AST with this PR merged to resolve https://github.com/rubocop-hq/rubocop/pull/8785?

lib/rubocop/ast/processed_source.rb
```diff
@@ -229,9 +229,9 @@ def parser_class(ruby_version)
         when 2.7
           require 'parser/ruby27'
           Parser::Ruby27
-        when 2.8
-          require 'parser/ruby28'
-          Parser::Ruby28
+        when 3.0
```
marcandre, line 232: could we have `when 2.8, 3.0` for compatibility?

marcandre:
And yes, I will make a release shortly, I would like to also merge 2 other PRs too.

lib/rubocop/ast/processed_source.rb
koic: Ah, I get it. In the future, the syntax of Ruby 2.8 and Ruby 3.0 may be dissociated, but I think they are almost the same now. It seems better that both are still supported as migration path. I updated this one.

marcandre: Well, this way we are compatible with older rubocop who know about 2.8 and not about 3.0. There won't be an actual Ruby 2.8 afaik.

marcandre:
Released 0.5.0 👍

koic:
Thank you for your quick releasing!

PR #128, opened 2020-09-28
[Fixes #126] Add JRuby and TruffleRuby to CI

### LINKED ISSUE #126: CI: check JRuby and other platforms
And avoid things like https://github.com/jruby/jruby/issues/6414 ...

marcandre:


spec/rubocop/ast/node_pattern_spec.rb
```diff
@@ -2062,10 +2062,12 @@ def withargs(foo, bar, qux)
     let(:ruby) { ':hello' }
     let(:instance) { defined_class.new }
 
-    if Set[1] === 1 # rubocop:disable Style/CaseEquality
-      let(:hello_matcher) { Set[:hello, :foo] }
-    else
-      let(:hello_matcher) { Set[:hello, :foo].method(:include?).to_proc }
+    let(:raise_argument_error) do
```
bquorning, line 2065: Instead of a `let`, you could make a custom matcher:
```rb
    RSpec::Matchers.define(:raise_argument_error) do
      supports_block_expectations

      match(notify_expectation_failures: true) do |proc|
        begin
          proc.call
        rescue ArgumentError => e
          expect(e.message).to include('wrong number of arguments')
          expect(e.backtrace_locations.first.lineno).to be(line_no) unless RUBY_PLATFORM == 'java'
        end
      end
    end
```
(the type check on `err` is make obsolete by `rescue ArgumentError`)

marcandre: I agree that the `let` (or a `def`) isn't pretty. Still, seems *much* simpler than a custom matcher, especially since I'm not expecting to reuse it much. Moreover, in your code it's not clear to me what happens if the `proc.call` succeeds and returns `true`...

marcandre: I'll merge this as is and it could be revised later on.

PR #131, opened 2020-09-29
Add rake tasks to merge and create Changelog entries

marcandre:
Fixes rubocop-hq/rubocop#8697

After yet another Changelog merge issue, I decided to write some code to avoid it.

This PR adds a rake task to merge entries in `CHANGELOG/` according to the prefix of their path: "new" (feature), (bug) "fix" or "change". It merges those with any existing entries in the unreleased section of the Changelog, so it can be done at any moment, is backward compatible, and nobody is forced to use it (except before cutting a release, and I've added a check for that 😅).

Other rake task `cl:new`, `cl:fix`, `cl:change` create an entry file (default use the last commit's message & git config).

For a contributor, the workflow looks like this:
* make a contribution, test it, etc.
* `git commit` (enter a good commit message, optionally starting with `[Fixes #xxx]`)
* `bundle exec rake cl:fix`
* modify file if desired
* `git commit --amend`
* no worry about any changelog conflict / forgetting to move it to the right spot if a release is made before it is merged

@bbatsov any objection that I use this in `rubocop-ast`? May I adapt it for `rubocop` too?

### DIFF
--- a/spec/tasks/changelog_spec.rb
+++ b/spec/tasks/changelog_spec.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+return unless RUBY_VERSION >= '2.6'
+
+require_relative '../../tasks/changelog'
+
+# rubocop:disable RSpec/ExampleLength
+RSpec.describe Changelog do
+  subject(:changelog) do
+    list = entries.to_h { |e| [e.path, e.content] }
+    described_class.new(content: <<~CHANGELOG, entries: list)
+      # Change log
+
+      ## master (unreleased)
+
+      ### New features
+
+      * [#bogus] Bogus feature
+      * [#bogus] Other bogus feature
+
+      ## 0.7.1 (2020-09-28)
+
+      ### Bug fixes
+
+      * [#127](https://github.com/rubocop-hq/rubocop-ast/pull/127): Fix dependency issue for JRuby. ([@marcandre][])
+
+      ## 0.7.0 (2020-09-27)
+
+      ### New features
+
+      * [#105](https://github.com/rubocop-hq/rubocop-ast/pull/105): `NodePattern` stuff...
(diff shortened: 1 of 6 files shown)


bbatsov:
> @bbatsov any objection that I use this in rubocop-ast? 

No objections.

> May I adapt it for rubocop too?

Sure. Might be good to use it for a while in `rubocop-ast` and refine the workflow (if needed) first, though.

tasks/changelog.rake
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+autoload :Changelog, "#{__dir__}/changelog"
+
+namespace :cl do
```
bbatsov, line 5: I think it's better to name this `:changelog` for clarity's sake. It's going to be easier to discover this way.

tasks/changelog.rb
```diff
@@ -0,0 +1,169 @@
+# frozen_string_literal: true
+
+if RUBY_VERSION < '2.6'
+  puts 'Changelog utilities available only for Ruby 2.6+'
+  exit(1)
+end
+
+# Changelog utility
+class Changelog
+  ENTRIES_PATH = 'CHANGELOG/'
```
bbatsov, line 10: I'd probably name the folder `changelog` instead, as I hate typing all caps. :D Names like `CHANGELOG` and `README` never sat well with me, but we're stuck with them for historical reasons. For the folder name we can be more flexible.

spec/tasks/changelog_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+return unless RUBY_VERSION >= '2.6'
+
+require_relative '../../tasks/changelog'
+
+# rubocop:disable RSpec/ExampleLength
+RSpec.describe Changelog do
+  subject(:changelog) do
+    list = entries.to_h { |e| [e.path, e.content] }
+    described_class.new(content: <<~CHANGELOG, entries: list)
```
bbatsov, line 11: I don't see tests making sure that the author links were added to the bottom of the changelog.

marcandre: I didn't implement that because I was lazy, but hey, why not... Done.

PR #132, opened 2020-09-29
Enable CI on JRuby

marcandre:


enebo:
@marcandre I noticed in GHA that JRuby is not run with --dev.  JRUBY_OPTS=--dev on my machine drops spec runs from 25s to 15s.  Hopefully takes a little sting out of adding us to CI.

enebo:
If set as an ENV this will also speed up the bundle install as well.

marcandre:
Cool, thanks @enebo! #175

enebo:
@marcandre I also submitted https://github.com/ruby/setup-ruby/pull/164 which should remove 10s of installing bundler on jruby-head (I noticed truffle ruby already does this and MRI 2.7+).

headius:
ruby/setup-ruby#164 has merged, perhaps we will see it in effect soon on rubocop-ast builds? 😀

marcandre:
Not yet, it is installing 9.2.16 right now.

bbatsov:
@marcandre I guess we should check if it's working now. 😅

headius:
FWIW, current JRuby runs both `spec` and `prism_spec` green.

marcandre:
Thanks

PR #134, opened 2020-09-30
[Fixes #78] Make numblock behave similarly to block with equivalent arguments.

### LINKED ISSUE #78: Handle numblock in each cops
Numbered Parameters feature has been introduced since 2.7.
The new node type, `numblock`, will be supported by rubocop-hq/rubocop#7665, but we still need to modify cops that use `on_block` or `block_type?` methods.

For example, `Style/BlockDelimiters` cop does not work for `numblock` even if the pull request is merged.

```ruby
# with TargetRubyVersion: 2.7, and the PR.

# The cop adds an offense to the following code.
items.each do |item| item / 5 end 

# The cop does not add an offense but it should.
items.each do _1 / 5 end
```

In this case, I guess we can add `alias on_numblock on_block` to the cop. And we can probably add the same code to the similar cops.

And I guess we also need to modify `Node#block_type?` method call.
I guess we can extend `numblock` node with `BlockNode` extension.

--------

## Expected behavior

Cops that work for `block` should work for `numblock` also.

## Actual behavior

It does not work well even with rubocop-hq/rubocop#7665 

## Steps to reproduce the problem

For `Stlye/BlockDelimiters` example, we can reproduce the problem with the following commands.

* Install RuboCop with rubocop-hq/rubocop#7665 
* Put .rubocop.yml with `TargetRubyVersion: 2.7`
* Put the example as a Ruby file
* Execute RuboCop

And we can find the potential target cops with git grep.

```bash
$ git grep on_block
lib/rubocop/ast/traversal.rb:      def on_block(node)
lib/rubocop/cop/layout/access_modifier_indentation.rb:        alias on_block  on_class
lib/rubocop/cop/layout/block_alignment.rb:        def on_block(node)
lib/rubocop/cop/layout/block_end_newline.rb:        def on_block(node)
lib/rubocop/cop/layout/empty_lines_around_access_modifier.rb:        def on_block(node)
lib/rubocop/cop/layout/empty_lines_around_block_body.rb:        def on_block(node)
lib/rubocop/cop/layout/indentation_width.rb:        def on_block(node)
lib/rubocop/cop/layout/multiline_block_layout.rb:        def on_block(node)
lib/rubocop/cop/layout/space_around_block_parameters.rb:        def on_block(node)
lib/rubocop/cop/layout/space_around_keyword.rb:        def on_block(node)
lib/rubocop/cop/layout/space_before_block_braces.rb:        def on_block(node)
lib/rubocop/cop/layout/space_inside_block_braces.rb:        def on_block(node)
lib/rubocop/cop/lint/next_without_accumulator.rb:        def on_block(node)
lib/rubocop/cop/lint/non_deterministic_require_order.rb:        def on_block(node)
lib/rubocop/cop/lint/redundant_with_index.rb:        def on_block(node)
lib/rubocop/cop/lint/redundant_with_object.rb:        def on_block(node)
lib/rubocop/cop/lint/useless_access_modifier.rb:        def on_block(node)
lib/rubocop/cop/lint/void.rb:        def on_block(node)
lib/rubocop/cop/metrics/block_length.rb:        def on_block(node)
lib/rubocop/cop/metrics/method_length.rb:        def on_block(node)
lib/rubocop/cop/mixin/method_complexity.rb:      def on_block(node)
lib/rubocop/cop/naming/block_parameter_name.rb:        def on_block(node)
lib/rubocop/cop/naming/variable_name.rb:        alias on_blockarg  on_lvasgn
lib/rubocop/cop/style/block_delimiters.rb:        def on_block(node)
lib/rubocop/cop/style/collection_methods.rb:        def on_block(node)
lib/rubocop/cop/style/each_for_simple_loop.rb:        def on_block(node)
lib/rubocop/cop/style/each_with_object.rb:        def on_block(node)
lib/rubocop/cop/style/empty_block_parameter.rb:        def on_block(node)
lib/rubocop/cop/style/empty_lambda_parameter.rb:        def on_block(node)
lib/rubocop/cop/style/for.rb:        def on_block(node)
lib/rubocop/cop/style/inverse_methods.rb:        def on_block(node)
lib/rubocop/cop/style/lambda.rb:        def on_block(node)
lib/rubocop/cop/style/method_called_on_do_end_block.rb:        def on_block(node)
lib/rubocop/cop/style/multiline_block_chain.rb:        def on_block(node)
lib/rubocop/cop/style/multiline_block_chain.rb:            # found by subsequent calls to on_block.
lib/rubocop/cop/style/next.rb:        def on_block(node)
lib/rubocop/cop/style/proc.rb:        def on_block(node)
lib/rubocop/cop/style/redundant_begin.rb:        def on_block(node)
lib/rubocop/cop/style/redundant_self.rb:        def on_blockarg(node)
lib/rubocop/cop/style/redundant_self.rb:        def on_block(node)
lib/rubocop/cop/style/redundant_sort_by.rb:        def on_block(node)
lib/rubocop/cop/style/single_line_block_params.rb:        def on_block(node)
lib/rubocop/cop/style/symbol_proc.rb:        def on_block(node)

$ git grep -F block_type?
lib/rubocop/ast/node/mixin/method_dispatch_node.rb:        parent&.block_type? && eql?(parent.send_node)
lib/rubocop/cop/lint/ambiguous_block_association.rb:          send_node.last_argument.block_type? &&
lib/rubocop/cop/lint/assignment_in_condition.rb:          return if node.condition.block_type?
lib/rubocop/cop/lint/shadowed_argument.rb:          node.conditional? || node.block_type? ||
lib/rubocop/cop/metrics/block_nesting.rb:          count_blocks? && node.block_type?
lib/rubocop/cop/mixin/multiline_expression_indentation.rb:          break false if a.block_type?
lib/rubocop/cop/mixin/multiline_expression_indentation.rb:          ancestor.block_type? && part_of_block_body?(candidate, ancestor)
lib/rubocop/cop/style/auto_resource_cleanup.rb:            (parent && (parent.block_type? || !parent.lvasgn_type?))
lib/rubocop/cop/style/inverse_methods.rb:          if node.block_type?
lib/rubocop/cop/style/method_call_with_args_parentheses/omit_parentheses.rb:            parent = node.parent&.block_type? ? node.parent.parent : node.parent
lib/rubocop/cop/style/method_called_on_do_end_block.rb:          return unless receiver&.block_type? &&
lib/rubocop/cop/style/multiline_block_chain.rb:            next unless receiver&.block_type? && receiver&.multiline?
lib/rubocop/cop/style/parallel_assignment.rb:          node.block_type? || node.send_type?
lib/rubocop/cop/style/redundant_parentheses.rb:          !ancestor.begin_type? && !ancestor.def_type? && !ancestor.block_type?
lib/rubocop/cop/style/safe_navigation.rb:          receiver = if method_chain.block_type?
lib/rubocop/cop/variable_force/variable.rb:          argument? && @scope.node.block_type?
lib/rubocop/cop/variable_force/variable_table.rb:            return nil unless scope.node.block_type?
lib/rubocop/cop/variable_force/variable_table.rb:            break variables unless scope.node.block_type?
lib/rubocop/cop/variable_force/variable_table.rb:          return unless current_scope.node.block_type?
spec/rubocop/ast/send_node_spec.rb:      it { expect(send_node.block_node.block_type?).to be(true) }
spec/rubocop/ast/super_node_spec.rb:      it { expect(super_node.block_node.block_type?).to be(true) }
```

## RuboCop version

master (with rubocop-hq/rubocop#7665 )

marcandre:
This aims to make it easier to deal the same way with `foo { |x, y| bar(x, y) }` than with `foo { bar(_1, _2) }`.

`parser` generates an AST that is different with regards to the arguments:

```rb
# for a block:
s(:args s(:arg :x) s(:arg :y))

# for a numblock:
2
```

The method `BlockNode#arguments` return the same nodes as one would have, had it been written as `do |_1, _1| ...`, except there are no locations.

The module `Traversal` will visit these "virtual" `arguments`, even though they are not actually part of the children.

`BlockNode#numbered_arguments` return `nil` for a normal block, and `2` for the example with `_1` and `_2`; probably not particularly useful

`BlockNode#arguments?` now returns `true` for a `numblock`

### DIFF
--- a/changelog/change_make_numblock_behave_similarly_to_block.md
+++ b/changelog/change_make_numblock_behave_similarly_to_block.md
@@ -1,0 +1,2 @@
+* [#78](https://github.com/rubocop-hq/rubocop-ast/issues/78): **(Potentially breaking)** `Block#arguments` now returns nodes for `numblock` (with no location) (Ruby 2.7+). ([@marcandre][])
+* [#78](https://github.com/rubocop-hq/rubocop-ast/issues/78): **(Potentially breaking)** `Traversal` now iterates on the arguments of `numblock` (Ruby 2.7+). ([@marcandre][])
--- a/changelog/fix_tweak_nodesource_so_it_does_not_crash_on.md
+++ b/changelog/fix_tweak_nodesource_so_it_does_not_crash_on.md
@@ -1,0 +1,1 @@
+* [#78](https://github.com/rubocop-hq/rubocop-ast/issues/78): Tweak `Node#source` so it does not crash on Nodes with no location. ([@marcandre][])
--- a/changelog/new_make_numblock_behave_similarly_to_block.md
+++ b/changelog/new_make_numblock_behave_similarly_to_block.md
@@ -1,0 +1,1 @@
+* [#78](https://github.com/rubocop-hq/rubocop-ast/issues/78): Add `Block#numbered_arguments` (Ruby 2.7+). ([@marcandre][])
--- a/lib/rubocop/ast/builder.rb
+++ b/lib/rubocop/ast/builder.rb
@@ -83,6 +83,20 @@ module RuboCop
         value(token)
       end
 
+      # @api private
+      # @return [Node] A node corresponding to the equivalent block `:args` node
+      def self.numblock_arguments(numblock)
+        nb = numblock.children[1]
+        type = nb == 1 && emit_procarg0 ? :procarg0 : :arg
+        args = 1.upto(nb).map do |i|
+          Node.new(type, [:"_#{i}"])
+        end
+        node = ArgsNode.new(:args, args)
+        node.send :parent=, numblock
+        node.complete!
+        node
+      end
+
       private
 
       def node_klass(type)
--- a/lib/rubocop/ast/node.rb
+++ b/lib/rubocop/ast/node.rb
@@ -238,9 +238,10 @@ module RuboCop
       end
 
       # Note: Some rare nodes may have no source, like `s(:args)` in `foo {}`
+      # or the `#arguments` of a `numblock`.
       # @return [String, nil]
       def source
-        loc.expression&.source
+        loc&.expression&.source
       end
 
       def source_range
--- a/lib/rubocop/ast/node/block_node.rb
+++ b/lib/rubocop/ast/node/block_node.rb
@@ -14,21 +14,27 @@ module RuboCop
       VOID_CONTEXT_METHODS = %i[each tap].freeze
       private_constant :VOID_CONTEXT_METHODS
 
+      def complete!
+        @mutable_attributes[:arguments] = Builder.numblock_arguments(self) if numblock_type?
+        super
+      end
+
       # The `send` node associated with this block.
       #
       # @return [SendNode] the `send` node associated with the `block` node
       def send_node
-        node_parts[0]
+        children[0]
       end
 
       # The arguments of this block.
+      # For `:numblock` type, simili arguments are returned
       #
       # @return [Array<Node>]
       def arguments
         if numblock_type?
-          [].freeze # Numbered parameters have no block arguments.
+          @mutable_attributes[:arguments]
         else
-          node_parts[1]
+          children[1]
         end
       end
 
@@ -36,7 +42,7 @@ module RuboCop
       #
       # @return [Node, nil] the body of the `block` node or `nil`
       def body
-        node_parts[2]
+        children[2]
       end
 
       # The name of the dispatched method as a symbol.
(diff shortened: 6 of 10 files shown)


marcandre:
cc/ @pocke @Drenmi @koic how does that sound?

I contemplated actually modifying the AST with our builder, which would have an impact for node patterns in particular, but I'm thinking we should probably have the same AST and children and use methods (like `arguments{?}` here) instead.

I was wondering if we should have a common `xblock_type?` like `call_type? == send_type? || csend_type?`, or if it's best to have patterns be more explicit with `{block numblock}`?

marcandre:
Just realized that there is `kwnilarg` which we don't currently handle which complicates matter a bit...

```ruby
def a(a=1, **nil)
end
def b(a=1)
end
a(foo: 42) # => error
b(foo: 42) # => ok, sets `a` to `{foo: 42}`
```

So I think I was confusing "has an argument signature" (`arguments{?}`) with "accepts arguments" (no existing method).

I'll close this PR and instead propose `accepts_arguments?` that takes these into account.

PR #141, opened 2020-10-22
Make `SendNode#macro?` aware of struct constructor

koic:
This PR makes `SendNode#macro?` aware of struct constructor.

Like class constructor, struct constructor will be recognized as a macro and will be awakened in the following `private` modifier:

```ruby
Foo = Struct.new(:foo) do
  private
    def private_foo
      foo
    end
end
```

With this change, this PR aims to resolve the following issue.
https://github.com/rubocop-hq/rubocop/issues/8919

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#141](https://github.com/rubocop-hq/rubocop-ast/pull/141): Make `SendNode#macro?` and `RuboCop::AST::Node#class_constructor?` aware of struct constructor and `RuboCop::AST::Node#struct_constructor?` is deprecated. ([@koic][])
+
 ## 1.0.0 (2020-10-21)
 
 ### Changes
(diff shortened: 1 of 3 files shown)


marcandre:
Very good. I consider that a bugfix.

Only thing I'm wondering... Should we not include `Struct.new` into `#class_constructor`?

I looked at the code in the main gem, and there is one `class_constructor? || struct_constructor?` and the only other use of `class_constructor?` (`Lint/DuplicateMethods`) should also include struct constructor.

I think it would be best to deprecate `struct_constructor?` and include that check in `class_constructor?`, or else create a `module_constructor?` that would include everything and deprecate both `class_constructor?` and `struct_constructor?`...

koic:
> I think it would be best to deprecate `struct_constructor?` and include that check in `class_constructor?`

Thank you for your review. I updated this PR with the suggestion.

bbatsov (approved):
I like the change.

marcandre:
Perfect, thanks!

marcandre:
Released in 1.0.1

koic:
@marcandre Wow! Thank you for the quick release ❤️

marcandre:
Of course. FWIW, I intend to release bug fixes within a day of merging. It's so easy to do a release and there's not that many PRs for `rubocop-ast`...

PR #142, opened 2020-10-22
Only traverse send nodes in def_modifier?

eugeneius:
A send node is a "def modifier" if it's part of a chain of method calls preceding a method definition. However the previous implementation was too permissive and matched any method definition in the send node's AST.

In the following example, `include` was considered to be a def modifier:

```ruby
include Module.new {
  private def foo
  end
}
```

This is the root cause of the problem fixed in https://github.com/rubocop-hq/rubocop/pull/8653.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -168,3 +168,4 @@
 [@fatkodima]: https://github.com/fatkodima
 [@koic]: https://github.com/koic
 [@dvandersluis]: https://github.com/dvandersluis
+[@eugeneius]: https://github.com/eugeneius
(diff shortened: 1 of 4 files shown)


marcandre:
Super!
Thanks for using the new changelog method. I'll make it clearer, but you didn't even have to enter your own nick's entry at the end of the Changelog.

eugeneius:
Oh, nice. I had originally added the changelog entry directly, but I moved it after reading CONTRIBUTING.md and left the link definition in place. I think the instructions are pretty clear, I'm just used to the old way.

marcandre:
👍  Perfect then.
I intend to make a release tomorrow btw.

marcandre:
Actually, I released 1.0.1 now :-)

PR #146, opened 2020-11-04
Fix `IfNode#branches` to return both branches when called on ternary conditional

fatkodima:
Currently, it is returning only 1 branch.

### DIFF
--- a/spec/rubocop/ast/if_node_spec.rb
+++ b/spec/rubocop/ast/if_node_spec.rb
@@ -434,6 +434,13 @@ RSpec.describe RuboCop::AST::IfNode do
       it { expect(if_node.branches).to all(be_literal) }
     end
 
+    context 'with a ternary operator' do
+      let(:source) { 'foo? ? :foo : 42' }
+
+      it { expect(if_node.branches.size).to eq(2) }
+      it { expect(if_node.branches).to all(be_literal) }
+    end
+
     context 'with an unless statement' do
       let(:source) { 'unless foo?; :bar; end' }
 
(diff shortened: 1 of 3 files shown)


marcandre:
Good catch, thanks.

Expect a release later today.

spec/rubocop/ast/if_node_spec.rb
```diff
@@ -434,6 +434,13 @@
       it { expect(if_node.branches).to all(be_literal) }
     end
 
+    context 'with a ternary operator' do
+      let(:source) { 'foo? ? :foo : 42' }
+
+      it { expect(if_node.branches.size).to eq(2) }
+      it { expect(if_node.branches).to all(be_literal) }
```
marcandre, line 441: FYI, I think you can do both in a single expectaction `to match [be_literal, be_literal]`

fatkodima: I just stick to the existing codestyle and replicated the nearest similar example 😄

marcandre:
Oh, didn't notice, but the Changelog entry was misplaced.

Best use `rake changelog:...` in the future!

marcandre:
Released 1.1.1, thanks!

fatkodima:
I do not understand, how I did not noticed that it is not under master? It is the second time, at least, that I remember. 💦

marcandre:
> I do not understand, how I did not noticed that it is not under master? It is the second time, at least, that I remember. 💦

You're probably too used to `rubocop` with a gazillion things in the "current" section

PR #150, opened 2020-11-11
Remove `:rasgn` and `:mrasgn` types.

marcandre:
See https://github.com/whitequark/parser/pull/739

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -3,7 +3,11 @@
 # NOTE: When changing minimal version of Ruby or Rubocop, change all of them
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 
 jobs:
   ast_specs:
--- a/docs/modules/ROOT/pages/node_types.adoc
+++ b/docs/modules/ROOT/pages/node_types.adoc
@@ -176,8 +176,6 @@ The following fields are given when relevant to nodes in the source code:
 
 |pair|One entry in a hash. |Two children, the key and value nodes.|1 => 2|https://rubydoc.info/github/rubocop-hq/rubocop-ast/RuboCop/AST/PairNode[PairNode]
 
-|rasgn|Right-hand assignment|Two children, the node representing the value to assign and the assignment node.|1 => a |
-
 |rational|Rational literal|One child, the Rational value|2.0r|N/A
 
 |redo|Redo command|None|redo|N/A
--- a/lib/rubocop/ast/node.rb
+++ b/lib/rubocop/ast/node.rb
@@ -237,7 +237,7 @@ module RuboCop
         each_ancestor.to_a
       end
 
-      # Note: Some rare nodes may have no source, like `s(:args)` in `foo {}`
+      # NOTE: Some rare nodes may have no source, like `s(:args)` in `foo {}`
       # @return [String, nil]
       def source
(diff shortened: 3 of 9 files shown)


marcandre:
@eregon Test runs hang on `truffleruby` (second time in a row). I stopped last run after 5 hours (I'm surprised github actions doesn't seem to have a cutoff?)

eregon:
It's possible to specify a timeout for GitHub Actions:
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

I'll try to reproduce. It seems to only affect `truffleruby-head`, right?

eregon:
It gets stuck in
https://github.com/rubocop-hq/rubocop-ast/blob/2788a1a0e129013f7e401a52b03a916b8dd0d495/lib/rubocop/ast/node_pattern/node.rb#L200-L204
Unfortunate that GitHub Actions doesn't send SIGINT or so, it would have shown that clearly.

TruffleRuby recently updated to 2.7, but didn't port that fix yet.

eregon:
This is quite involved because I had to finish writing a spec for Range#minmax, synchronize specs across implementations and then actually fix it in TruffleRuby.

eregon:
BTW, is there any reason not to use `Range#begin/end` here? It's always very fragile to depends on `RUBY_VERSION` checks.

marcandre:
Thanks for the investigation!

> is there any reason not to simply use `Range#begin/end` here

I don't remember, and it looks like all the arity ranges are inclusive, so maybe not. Still, I'd rather write things properly; also I should probably be using refinements here. Does `map(&:refined_method)` work in truffleruby?

>  It's very fragile to depend on `RUBY_VERSION` checks.

Agreed, it's also not the best in general. I think I can use a different check like `(0..4.2).minmax.last`

eregon:
Fixed in https://github.com/oracle/truffleruby/commit/df78798c151e87ae79e1e882cb5c64f70e01d1e6, thanks for the ping.
So no need to change anything :)

> I don't remember, and it looks like all the arity ranges are inclusive, so maybe not.

Exclusive ranges wouldn't work anyway, 2.7.2:
```
> (0...Float::INFINITY).minmax
TypeError: cannot exclude non Integer end value
```

> Does map(&:refined_method) work in truffleruby?

It does. But implementation-wise, that's a proper nightmare (need to cache globally per execution context & per refinements), I'd recommend having an explicit block.

marcandre:
> Fixed in [oracle/truffleruby@df78798](https://github.com/oracle/truffleruby/commit/df78798c151e87ae79e1e882cb5c64f70e01d1e6), thanks for the ping.
> So no need to change anything :)
> 
> > I don't remember, and it looks like all the arity ranges are inclusive, so maybe not.
> 
> Exclusive ranges wouldn't work anyway, 2.7.2:

Right. For `parser`'s TreeRewriter, I break the integers into exclusive integer ranges and a last inifnite range (that I make inclusive because it doesn't work otherwise, but I think that's unfortunate): `0...15, 15...42, 42..Float::INFINITY`. Works like a charm.

> > Does map(&:refined_method) work in truffleruby?
> 
> It does. But implementation-wise, that's a proper nightmare (need to cache globally per execution context & per refinements), I'd recommend having an explicit block.

Ok. Since TruffleRuby is fixed now, I could use a conditional refinement here to support MRI <= 2.6. Should be nicer.

PR #154, opened 2020-11-23
Allow `ArgsNode` to iterate over its arguments

dvandersluis:
While working on rubocop-hq/rubocop#9071, I needed a way to evaluate block arguments (and get their names) that are inside destructuring, but `BlockNode#arguments` returns the `mlhs` node instead of the actual arguments. This change adds `each_arg` to `ArgsNode` so that only argument types can be iterated, and created a Node type (`ArgNode`) for the 7 arg types, which exposes `name`, `default_value`, etc. I added some predicates that I think are useful too.

I overrode `ArgsNode#to_a` so that `arguments.size` will return the actual number of arguments when they are not all at the same node depth.

Finally, this adds `BlockNode#argument_names` which is convenient for me in rubocop-hq/rubocop#9071. It gives a way to collect all the argument names even when there are numbered params (for which `BlockNode#arguments` just returns a bare array).


lib/rubocop/ast/builder.rb
```diff
@@ -20,6 +20,13 @@ class Builder < Parser::Builders::Default
       NODE_MAP = {
         and:          AndNode,
         alias:        AliasNode,
+        arg:          ArgNode,
+        blockarg:     ArgNode,
+        kwarg:        ArgNode,
+        kwoptarg:     ArgNode,
+        kwrestarg:    ArgNode,
+        optarg:       ArgNode,
+        restarg:      ArgNode,
```
iliabylich, line 29: there's also a `shadowarg`

lib/rubocop/ast/node/args_node.rb
```diff
@@ -24,6 +24,45 @@ class ArgsNode < Node
...
+      # @overload each_arg(type_a, type_b, ...)
+      #   Yield only arguments matching any of the types.
+      # @yieldparam [Node] node each arg
+      # @return [self] if a block is given
+      # @return [Enumerator] if no block is given
+      def each_arg(*types, &block)
+        if types.empty?
+          types = ARGUMENT_TYPES.dup
+          types << :forward_arg if RuboCop::AST::Builder.emit_forward_arg
+        end
```
dvandersluis, line 47: @marcandre I'm not sure if `ARGUMENT_TYPES` should include `forward_arg`?

lib/rubocop/ast/node/arg_node.rb
```diff
@@ -0,0 +1,55 @@
...
+  module AST
+    # A node extension for `arg`, `optarg`, `restarg`, `kwarg`, `kwoptarg`,
+    # `kwrestarg` and `blockarg` nodes. This will be used in place of a plain
+    # node when the builder constructs the AST, making its methods available
+    # to all `arg` nodes within RuboCop.
+    class ArgNode < Node
+      # Returns the name of an argument.
+      #
+      # @return [Symbol] the name of the argument
+      def name
```
iliabylich, line 13: it can return `nil` for `def m(*); end` and `def m(**); end`

lib/rubocop/ast/builder.rb
dvandersluis: Can you give me an example of it?

dvandersluis: Never mind [found it](https://github.com/whitequark/parser/blob/master/doc/AST_FORMAT.md#block-shadow-arguments). Same question then as `forward_arg`, should it be added to the `Node::ARGUMENT_TYPES` set?

marcandre:
Thank you for the great PR. I meant to do something like that for a while too. I'm definitely 👍 for the creation of `ArgNode`. 

One thing to keep in mind is `numblock`. There's a distinction between the number of children in the AST node that isn't always clean. I had [started a PR](https://github.com/marcandre/rubocop-ast/commit/814103c9e1d4ed2c334a98171ae2e448d7181594) but then realized that my implementation was not a good idea. Anyways, it is not necessary for this PR though, but we must keep that in mind.

As far as the interface goes, I'd prefer to keep `to_a` intact and provide a new method to access the information.
Instead of `each_arg`, and `names` how about only `argument_list`? We could also make it accessible from `BlockNode` + `NumblockNode`.

lib/rubocop/ast/node/arg_node.rb
marcandre: and `...`

dvandersluis:
@marcandre sounds good about `argument_list`, I'll make that change. The only issue there IMO is that for `foo { |x, (y, z)| }`, `arguments.size` will still be 2, but `arguments_list.size` will be 3 so that's okay.

lib/rubocop/ast/node/arg_node.rb
dvandersluis: These will just return nil already but I’m adding tests for them.

lib/rubocop/ast/builder.rb
marcandre: I always forget about these. I've never used them, or seen them in the wild!

> Same question then as `forward_arg`, should it be added to the `Node::ARGUMENT_TYPES` set?

Yes, it should.

lib/rubocop/ast/node/args_node.rb
marcandre: Yes, please include it

marcandre:
> @marcandre sounds good about `argument_list`, I'll make that change. The only issue there IMO is that for `foo { |x, (y, z)| }`, `arguments.size` will still be 2, but `arguments_list.size` will be 3 so that's okay.

Exactly. And I'm thinking `{ _1 + _2 }` should have `arguments.size` be 0, but `arguments?` be `true` and `argument_list.size` be 2.

dvandersluis:
@marcandre updated to use `arguments_list` instead of the previous solution.

I'm not really sure I understand what a `procarg0` is, but it seems a block with 1 argument is parsed to it with modernize mode, so I added a class for it as well. It couldn't just reuse `ArgNode` because the `arg` node is its child, so `name` is overridden. I think we probably can do something similar for `ForwardArgsNode`, but since it's a collection I didn't want to change it in case I was overlooking something. Instead, I made the specs that may parse to a `forwards_arg` conditional on `RuboCop::AST::Builder.emit_forward_arg`.

> And I'm thinking { _1 + _2 } should have arguments.size be 0, but maybe arguments? should be true and argument_list.size be 2.

I didn't do this for this PR but I'll try to get this working next.

marcandre:
> I'm not really sure I understand what a `procarg0`

It is to distinguish `foo {|x| ... }` and `foo {|x,| ... }` which behave a bit differently and were parsed the same way. That distinction rarely affects `rubocop` so we still use the legacy format by default.

> so I added a class for it as well.

Perfect. It's a valid question as to why the format is like this though, I'll ask.

changelog/new_add_blocknodeargument_names.md
```diff
@@ -0,0 +1,2 @@
+* [#154](https://github.com/rubocop-hq/rubocop-ast/pull/154): Add `ArgNode` and `Procarg0Node`, and add `ArgsNode#argument_list` to get only argument type nodes. ([@dvandersluis][])
```
marcandre, line 1: `Procarg0Node` => `Procarg0Node` ("Modern" mode)

lib/rubocop/ast/node/arg_node.rb
```diff
@@ -0,0 +1,70 @@
...
+        restarg_type?
+      end
+
+      # Checks whether the argument is double splatted.
+      #
+      # @return [Boolean] whether the argument is a keyword double splat
+      def double_splatted?
+        kwrestarg_type?
+      end
+      alias keyword? double_splatted?
```
marcandre, line 53: I don't understand `keyword?` here... What about `kwarg` and `kwoptarg`?

lib/rubocop/ast/node/block_node.rb
```diff
@@ -32,6 +35,23 @@ def arguments
...
+      def argument_list
+        return [] unless arguments?
+
+        arguments.argument_list
+      end
+
+      # The argument names of this block.
+      #
+      # @return [Array<Symbol>]
+      def argument_names
```
marcandre, line 51: I'm not sure this method is needed. Let's remove it for now and if we keep using it we can create a shortcut

lib/rubocop/ast/node/arg_node.rb
```diff
@@ -0,0 +1,70 @@
...
+      def default_value
+        return unless optional?
+
+        node_parts[1]
+      end
+
+      # Checks whether the argument is optional (it has a default value)
+      #
+      # @return [Boolean] whether the argument is optional
+      def optional?
```
marcandre, line 29: I'm not sure this is the right name, or the right result.
Maybe it should be named `default?`
Or if `optional?`, then I'd expect `argument_list.all?(&:optional?)` to be a good check if there's a mandatory argument, so I'd expect a `true` result for `*x`, `**x`, `...` and `&block`.
Or maybe you want both 😆

lib/rubocop/ast/node/arg_node.rb
```diff
@@ -0,0 +1,70 @@
...
+        optarg_type? || kwoptarg_type?
+      end
+
+      # Checks whether the argument is a block argument
+      #
+      # @return [Boolean] whether the argument is a block argument
+      def block?
```
marcandre, line 36: I'm not sure that these aliases add much in terms of clarity. I realize you put lots of work in the tests, but I'm not convinced we need them and I fear they just add to the cognitive load (a reader might wonder... is that different from `blockarg_type?`?)

spec/rubocop/ast/block_node_spec.rb
```diff
@@ -43,6 +49,54 @@
...
+      let(:source) { 'foo { |a, b = 42, (c, *d), e:, f: 42, **g, &h| nil }' }
+
+      it { expect(argument_list.size).to eq(8) }
+    end
+
+    context '>= Ruby 2.7', :ruby27 do
+      context 'using numbered parameters' do
+        let(:source) { 'foo { _1 }' }
+
+        it { is_expected.to be_empty }
```
marcandre, line 71: I hope this will change...

marcandre (changes requested):
I like where this is going!

lib/rubocop/ast/node/arg_node.rb
dvandersluis: Yeah I think I was mid-thought and got distracted.

lib/rubocop/ast/node/block_node.rb
dvandersluis: Yeah I was hoping to squeak it in because I want it in my rubocop PR but that sounds reasonable 😅

lib/rubocop/ast/node/arg_node.rb
dvandersluis: I was going based on the nodes being `optarg` and `kwoptarg`, but I think you have a valid point. I'll change it to `default?`.

lib/rubocop/ast/node/arg_node.rb
dvandersluis: Eh I'm not worried about removing things 😄 I was trying to give this a fluent interface, but the ones that are just essentially `*_type` aliases are probably not useful. I'll remove `block?`, `splatted?`, `double_splatted?` (which I didn't like as a name, hence the half thought out alias 😅) `shadow?` and `forwarding?`.

spec/rubocop/ast/block_node_spec.rb
dvandersluis: It does in #155!

dvandersluis:
@marcandre I wasn't sure if `procarg0` should be added to node_types.adoc? I don't think any of the "modern" node types are documented there, but now we have a class for it.

marcandre (commented):
Looks pretty much ready to merge.
One last small detail, could we return frozen arrays? We might want to cache this in the future, and there shouldn't be a need to modify this result.

lib/rubocop/ast/node/block_node.rb
```diff
@@ -32,6 +35,16 @@ def arguments
         end
       end
 
+      # Returns a collection of all descendants of this node that are
+      # argument type nodes. See `ArgsNode#argument_list` for details.
+      #
+      # @return [Array<Node>]
+      def argument_list
+        return [] unless arguments?
```
dvandersluis, line 43: @marcandre is there an array other than this one that you want me to freeze? Not seeing anything else.

dvandersluis: Actually I think you meant `arguments_list` so I've frozen that as well now.

marcandre:
Great work, thanks!

PR #155, opened 2020-11-23
Enable `BlockNode#argument_list` for `numblock`s. 

dvandersluis:
Follows #154 (I'll rebase away that commit once merged).

`numblock`s now return an array of `ArgNode`s from `BlockNode#argument_list`, ~and `BlockNode#arguments?` is true for `numblock`s.~

### DIFF
--- a/spec/rubocop/ast/block_node_spec.rb
+++ b/spec/rubocop/ast/block_node_spec.rb
@@ -52,6 +52,8 @@ RSpec.describe RuboCop::AST::BlockNode do
   describe '#argument_list' do
     subject(:argument_list) { block_node.argument_list }
 
+    let(:names) { argument_list.map(&:name) }
+
     context 'with no arguments' do
       let(:source) { 'foo { bar }' }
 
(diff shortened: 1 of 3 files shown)


dvandersluis:
@marcandre you mentioned that maybe `arguments?` should be true for a numblock but now I think this is a bad idea. A whole bunch of cops assume `arguments?` mean that there are explicit arguments, and this will break expectations. I'm going to revert that part.

marcandre:
Sounds good. I'm curious if all of these errors are genuine, or if some cops are just not `numblock` aware.

Maybe we should have a method `accepts_arguments?` (or `arity?` or something)?

dvandersluis:
I think maybe most methods are not `numblock` aware. There was one actual test failure, but then when I was looking at it, I noticed a lot of places use `arguments?` to determine how to autocorrect (which means that if `numblock` arguments count, we'll get some weird autocorrect, which is why I thought we should leave it as is).

Out of curiosity, what's the procedure for changing something that breaks rubocop's tests?

marcandre:
> Out of curiosity, what's the procedure for changing something that breaks rubocop's tests?

Ideally we make the main gem work first, then we can bump the minimum main gem we fully support in the CI specs and builds are always good...

dvandersluis:
@marcandre this is rebased now.

spec/rubocop/ast/block_node_spec.rb
```diff
@@ -66,9 +66,17 @@
...
+          let(:source) { 'foo { _7 }' }
+
+          it { expect(argument_list.size).to eq(7) }
+        end
+
+        context 'with sequential params' do
+          let(:source) { 'foo { _1 + _2 }' }
 
-        it { is_expected.to be_empty }
+          it { expect(argument_list.size).to eq(2) }
```
marcandre, line 78: Could you test the names instead of just the size please?

dvandersluis: Done!

spec/rubocop/ast/block_node_spec.rb
```diff
@@ -52,25 +52,37 @@
...
+
+        context 'with sequential params' do
+          let(:source) { 'foo { _1 + _2 }' }
 
-        it { is_expected.to be_empty }
+          it { expect(names).to eq(%i[_1 _2]) }
+        end
       end
     end
+    # rubocop:enable Naming/VariableNumber
```
marcandre, line 85: I opened https://github.com/rubocop-hq/rubocop/pull/9091...

marcandre:
Released as 1.2.0, thanks! 👍

dvandersluis:
Thanks @marcandre! ❤️

PR #157, opened 2020-11-30
NodePattern now considers constant names to refer to constants

### LINKED ISSUE #156: NodePattern for numblock "count" element
Consider:

```shell
ruby-parse --27 -e "something { _1 }"
(numblock
  (send nil :something) 1
  (lvar :_1))
```

The count of numbered parameters is returned as a bare integer, which makes it extra resistant to being captured in a node pattern. It's not an `int` node so `int` or `int_type?` don't work, and if I try `#is_a?(Integer)` I get `ArgumentError: wrong number of arguments (given 2, expected 1)` for some reason.

Any suggestions?

marcandre:
(instead of predicate `#Example_type?`)

[Fixes #156]

### DIFF
--- a/changelog/fix_nodepattern_now_considers_constant_names.md
+++ b/changelog/fix_nodepattern_now_considers_constant_names.md
@@ -1,0 +1,1 @@
+* [#156](https://github.com/rubocop-hq/rubocop-ast/issues/156): NodePattern now considers constant names to refer to constants (instead of predicate `#Example_type?`). ([@marcandre][])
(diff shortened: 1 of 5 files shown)


marcandre:
Another difference is that this PR prevents naming unification variables with a first letter that is a capital `(int _Example)`.

marcandre:
Noticed that identifiers where really too lax, for example `#_foo` or `bar-baz?` are better not be valid

PR #163, opened 2020-12-29
Configure CI for ruby 3.0

tejasbubane:
https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -23,14 +23,14 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu ]
-        ruby: [ 2.4, 2.5, 2.6, 2.7, head ]
+        ruby: [ 2.4, 2.5, 2.6, 2.7, 3.0, head ]
         coverage: [ null ]
         modern: [ null ]
         title: [ null ]
         include:
           - { os: windows, ruby: mingw }
           - { ruby: 2.4, os: ubuntu, coverage: true, title: 'Coverage' }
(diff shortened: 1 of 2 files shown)


.github/workflows/rubocop.yml
```diff
@@ -81,7 +81,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu ]
-        ruby: [ 2.4, 2.7 ]
+        ruby: [ 2.4, 3.0 ]
         rubocop: [ 'v0.92.0', master ]
```
marcandre, line 85: Ruby 3.0 & v0.92 are an issue because of `SafeYAML`. This was resolved in v1.3.0, so maybe we could test v0.92.0 in Ruby 2.4, and 1.3.0 in Ruby 2.4 and 3.0?

bbatsov: Why do we even care about 0.92 at this point?

marcandre: True, we don't really need to, we could also simply bump the minimal rubocop version, and we have a good mechanism to do so now.

koic: The version specification in YAML should be string `'3.0'` instead of float `3.0`. It will be `3` instead of the expected `3.0`.

result-of-ci

marcandre:
I enabled the main specs for Ruby 3.0 in master only as there was a bunch of builds that today have issues because of regex_parser & utf-8 changes, and even last rubocop build was broken (bug I did in changelog script 😅).

PR #167, opened 2021-01-23
Fix `#value` for a `dstr` node.

dvandersluis:
`dstr` nodes previously used `StrNode`, which exposes `#value`; however, it just returns the first child, which is a node, not a string value.

This changes `dstr`s to have their own node class that inherits from `StrNode`  but overrides `#value` as to get the actual  string value of the node (including any interpolations).

ie.
```ruby
'foo' \
'bar'

# previously:
dstr_node.value => s(:str, "foo")

# after change:
dstr_node.value => 'foobar'
```


marcandre:
Seems like a good addition. Would you add a Changelog entry (same way as the main gem) please?

dvandersluis:
Definitely!

marcandre:
Still can't see it...

dvandersluis:
@marcandre updated now!

marcandre:
Thanks!
Released 1.4.1

tas50:
@dvandersluis Thanks for fixing this. I had actually been working around it.

PR #168, opened 2021-01-28
Replace incorrect node_matcher name

adilw3nomad:
I was looking at the docs today and noticed that the section for arguments defines a node matcher as `has_sensitive_data?`, however it then goes on to use `has_user_data?` in examples.


marcandre:
Good catch, thanks!

PR #171, opened 2021-03-08
Simplify `def_modifier`. Improve doc.

marcandre:


marcandre:
@bbatsov Should I create a different method without the `?`?

marcandre:
Or maybe @dvandersluis ?

dvandersluis:
I don't have a strong opinion here but it probably makes more sense to not be a predicate method if its return method is purposefully an actual object, otherwise it's kind of a side effect.

bbatsov:
I agree. I'd introduce a new method named `def_modifier` with the proposed behaviour. I've never been found of predicate methods that don't return booleans.

marcandre:
Ok, PR changed accordingly 👍

PR #176, opened 2021-04-04
Fix unexpected non-local exits on Rubocop::AST::Node#parent_module_name

sanfrecce-osaka:
With this PR, when Rubocop::AST::Node#parent_module_name_part is executed, it will no longer do non-local exits when a block is evaluated, and evaluating singleton_class or block in Rubocop::AST::Node#parent_module_name will behave as expected.
With this change, this PR aims to resolve the following issue.

https://github.com/rubocop/rubocop/issues/5022


marcandre:
Thanks for the PR.

I can only look at this more fully tomorrow, but this PR is not acceptable as is. The original intent was that `parent_module_name_part` could either return or yield a value (potentially `nil`). This PR now conflates both and result is over complicated for what it achieves.

Either `parent_module_name_part` should simply return a value, or else it needs to be modified in some way to achieve what we need.

marcandre:
Closed in favor of #177

marcandre:
See #197

PR #177, opened 2021-04-05
Improve `parent_module_name`

marcandre:
This builds on #176 

@sanfrecce-osaka: first commit shows a way to simplify the tests

Second commit adds a test that would fail with your patch, and fixes that.

This still isn't quite right as the result with `duplicate_method` cop isn't quite right, and it gets confused between `Foo.method` and `Foo::#<Class:Foo>#method`.

I'm not sure if it's best to change the cop, or add an `Node` method for `qualified_name` or something.


bbatsov:
I'm fine with both proposals, although I'm leaning slightly towards the second option.

marcandre:
Merging as is, further improvements welcome

marcandre:
I'm not sure how come the test suite was green. This deteriorates even some simple cases. Reverted.

PR #185, opened 2021-05-28
Fix doc url to Rubydoc

lulalala:
After the namespace upgrade from "rubocop-hq" to "rubocop", the links to rubydoc are broken as rubydoc dose not offer redirection.

I didn't rename github urls since they work currently.


bbatsov:
Thanks!

PR #186, opened 2021-05-28
Add `pattern` method for `AST::InPatternNode`

koic:
This PR adds `pattern` method for `AST::InPatternNode`.


marcandre:
Thanks! Release in 1.7.0 👍

PR #193, opened 2021-08-02
Fix typo iff -> if

brunoarueira:


marcandre:
Not [actually a typo](https://en.wikipedia.org/wiki/If_and_only_if). Rewording as "returns wether ..."

brunoarueira:
@marcandre great TIL 😅

PR #195, opened 2021-08-05
Move `ProcessedSource#sorted_tokens` to be a public method

dvandersluis:
A number of cops are not detecting issues when looking at tokens around heredocs (for instance `Layout/SpaceInsideParams`) because they use pairs of tokens, but the tokens are not sorted properly for heredocs as method arguments (as per the comment of `ProcessedSource#sorted_tokens`).

For instance, this space is not detected by `Layout/SpaceInsideParams` currently because the tokens are not sorted in the expected order:

```ruby
foo(<<~STR )
  text
STR
```

This change just moves that method to be public so that the cops can use it.

### DIFF
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -177,6 +177,14 @@ module RuboCop
         sorted_tokens[last_token_index(range_or_node)]
       end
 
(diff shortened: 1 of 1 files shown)


marcandre:
Released in 1.9.0 👍

dvandersluis:
Thank you!

PR #197, opened 2021-08-10
[Fix #184] Fix `Node#parent_module_name` for `sclass` nodes

### LINKED ISSUE #184: Issues with `parent_module_name`
Start from #177

dvandersluis:
Continuation of #177. I took @marcandre's tests and got them to pass. This fixes rubocop/rubocop#5022, but rubocop itself needs a PR as well because this change adds elements to the `parent_module_name` that it is not currently expecting.

### DIFF
--- a/changelog/fix_fix_184_fix_nodeparent_module_name_for.md
+++ b/changelog/fix_fix_184_fix_nodeparent_module_name_for.md
@@ -1,0 +1,1 @@
+* [#197](https://github.com/rubocop-hq/rubocop-ast/pull/197): [Fix #184] Fix `Node#parent_module_name` for `sclass` nodes. ([@dvandersluis][])
--- a/lib/rubocop/ast/node.rb
+++ b/lib/rubocop/ast/node.rb
@@ -317,7 +317,11 @@ module RuboCop
         # returns nil if answer cannot be determined
         ancestors = each_ancestor(:class, :module, :sclass, :casgn, :block)
         result    = ancestors.map do |ancestor|
-          parent_module_name_part(ancestor) { |full_name| return full_name }
+          parent_module_name_part(ancestor) do |full_name|
+            return nil unless full_name
+
+            full_name
+          end
         end.compact.reverse.join('::')
         result.empty? ? 'Object' : result
       end
--- a/spec/rubocop/ast/node_spec.rb
+++ b/spec/rubocop/ast/node_spec.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::AST::Node do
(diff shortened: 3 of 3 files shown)


dvandersluis:
There's a bit of a catch 22 between this PR and rubocop/rubocop#9994. Each PR needs the other to get the tests passing.

dvandersluis:
There's a bit of a catch 22 between this PR and rubocop/rubocop-ast#197. Each PR needs the other to get the tests passing.

marcandre:
> There's a bit of a catch 22 between this PR and [rubocop/rubocop#9994](https://github.com/rubocop/rubocop/pull/9994). Each PR needs the other to get the tests passing.

Right. So either we merge this and reset the CI suite to test from the next version of RuboCop, or else we introduce some flag.

Let me see how your additional tests fail with the previous code

marcandre:
Yeah, no flag.

So, short of introducing a list of tests that are ok to fail, I think the only way forward is:
1) merge this
2) tweak the CI suite by removing tests on RuboCop 0.92
3) release a bug fix release
On RuboCop's side:
4) merge PR
5) bump the minimum requirement for `rubocop-ast`
6) release a version of RuboCop
Back on RuboCop::AST:
7) tweak the CI suite to restore testing on that version of RuboCop

All RuboCop PRs between step 3 and 4 will have their CI fail though.

@dvandersluis do you see a better way?

dvandersluis:
So like you said the problem is that this change will break existing tests in rubocop, even without the new tests, because the method return is going to be different.

I'm good with removing rubocop 0.92 from the test suite, since I'm not really sure why we're running against a really old rubocop anyways? However, if we need it we can always add it back when done (or maybe a more recent version?).

I think maybe a less disruptive path would be:
1. Remove rubocop 0.92 from CI here, as discussed
2. On rubocop master: mark all the tests that would fail with this change as pending
3. Merge this in
4. Revert the changes from (2) and bump the rubocop-ast version simultaneously.
5. Update rubocop-ast CI.

If we agree with this (and are good with both this and the rubocop PR, obviously) I can do all this quickly and sequentially for minimal disruption.

marcandre:
> we're running against a really old rubocop anyways

Ideally, projects with old versions of RuboCop (e.g. stuck with old versions of Ruby) can still run with newer `rubocop-ast`.

> If we agree with this

Indeed, marking the tests as pending is the way to go 👍 Let's do this.

I'll be glad to make the release. If you create the PR to tweak the CI suite, don't forget to modify the `.mergify` too

dvandersluis:
@marcandre once #198 is merged, I'll rebase and merge this PR.

dvandersluis:
@marcandre this is ready to be merged and released now.

marcandre:
Released 1.9.1, thanks! 👍

PR #201, opened 2021-08-12
Add Node classes for assignment types

dvandersluis:
Adds nodes with a mostly unified interface for `lvasgn`, `ivasgn`, `cvasgn`, `gvasgn`, `casgn`, `op-asgn`, `and-asgn` and `or-asgn` nodes.

The idea here is that we have a lot of repeated boilerplate in some cops to handle assignment nodes that are mostly but not completely the same. This will allow all assignment nodes to respond to `name` and `expression` regardless of the node type.

I didn't create a node class for `masgn` because that works differently than everything else and I'm not exactly sure what we'd want there as of yet.


dvandersluis:
Once this is available I have some cops I'd like to update to make more simple. 😄

marcandre:
Great work 🎉

PR is great as is, but could you also update `docs/modules/ROOT/pages/node_types.adoc` please?

marcandre (approved):

dvandersluis:
@marcandre absolutely! I actually had that file open to use as a reference but forgot to update the nodes 😅 Fixed now!

marcandre:
Released 1.10.1. Thanks!

PR #202, opened 2021-08-12
Remove unrelated comments from `ConstNode`

dvandersluis:
Looks like `ConstNode` was copied from `BlockNode` and left some unrelated comments.


marcandre:
Thanks 👍

PR #205, opened 2021-08-24
Make class, module, and struct definitions aware of numblock

koic:
This PR makes class, module, and struct definitions aware of numblock.


marcandre:
Looks good.

If this comes up a lot, we should consider supporting `(gen_block ...)` or something...

marcandre:
Released 1.11.0

koic:
Great! Thanks!

PR #210, opened 2021-09-23
Update `hash-pattern` to use `HashNode` and `array-pattern` to use `ArrayNode`.

dvandersluis:
`hash-pattern` and `array-pattern` are special case hashes and arrays so should be able to be used like normal.


dvandersluis:
The `Lint/RequireRelativeSelfPath` failures will be fixed by rubocop/rubocop#10121

dvandersluis:
I think the ruby-head failure is related to https://bugs.ruby-lang.org/issues/18177 @koic?

dvandersluis:
The previous CI failures are all green now!

marcandre:
> `hash-pattern` and `array-pattern` are special case hashes and arrays

I am not convinced of that. Can you tell me more why you think so, and what use-cases you have in mind?

I'm not against extracting common methods, but we have to think carefully about this.

Although it looks similar, pattern match means something very different.

E.g.: `{x: }` means something as a pattern match, it also does not map really well to `each_key` / `each_pair`. Moreover, in Ruby 3.1, it will mean something completely different when not a pattern match.

dvandersluis:
Thanks for reviewing @marcandre!

I had opened this for https://github.com/rubocop/rubocop/pull/10108, but ended up solving it another way. My rationale is that `array-pattern` and `hash-pattern` are hashes that just happen to contain `match-var`s instead of other nodes. Fair point about `each_key` and `each_pair` though, and I wasn't thinking about how it could match a hash key without capturing a value from it. 

I don't actively need this anymore so if it doesn't make sense feel free to close it, or I'm happy to make changes to make this make more sense.

marcandre:
Let's close this until we actually need to add functionality to the pattern nodes

PR #213, opened 2021-11-06
Make `Node#numeric_type?` aware of rational and complex literals

koic:
This PR makes `Node#numeric_type?` aware of complex and rational literals.
These literals could be treated as numeric.

```ruby
42r.is_a?(Numeric)  # => true
42i.is_a?(Numeric)  # => true
42ri.is_a?(Numeric) # => true
```


marcandre:
Thanks! Released in 1.13.0

PR #216, opened 2021-11-12
Set `rubygems_mfa_required` in gemspec

dvandersluis:
Follows rubocop/rubocop#10239. Update gemspec to require MFA for privileged operations.


marcandre:
Thanks!
I'm wondering if I should I make a release?

dvandersluis:
Probably wouldn't hurt but also if we just wait til the next actual code change that's probably okay too!

PR #219, opened 2021-12-12
Add `value_omission` method to `AST::PairNode`

koic:
Follow up to https://github.com/whitequark/parser/pull/818

This PR adds `value_omission` method to `AST::PairNode` for Ruby 3.1's hash value omission.
It can replace string match with that named API.


koic:
The CI failure is truffleruby-head, so I think the matrixes required for the merge has passed.

marcandre:
Thank you!
Released in v1.15

koic:
Thanks!

PR #220, opened 2021-12-27
Make `AST::Node#receiver` aware of `csend` block method calls

koic:
Resolves: https://github.com/rubocop/rubocop/issues/10319.

This PR makes the following behavior of `send` calls and `csend` calls consistent.

Before:

- The receiver of `foo.bar { baz }` is `foo`.
- The receiver of `foo&.bar { baz }` is `nil`.

After:

- The receiver of `foo.bar { baz }` is `foo`.
- The receiver of `foo&.bar { baz }` is `foo`.


koic:
The TruffleRuby CI matrix is failing, but it probably is unrelated to this change.
https://github.com/rubocop/rubocop-ast/runs/4639108876

marcandre:
Absolutely 👍

marcandre:
Released as 1.15.1, thanks!

PR #232, opened 2022-04-26
Drop Ruby 2.5 support

koic:
Follow up https://github.com/rubocop/rubocop/pull/10577.


marcandre:
Thank you 👍

I assume no release is needed

PR #233, opened 2022-05-13
Make parse from Ruby 1.9 to 2.3 available

koic:
RuboCop can analyze Ruby 1.9 or later. But now it turns out that RuboCop hasn't been able to parse older Ruby versions.
https://docs.rubocop.org/rubocop/1.29/compatibility.html#support-matrix

Analysis Ruby code and runtime Ruby versions are different. So, only runtime version should have been dropped.

This PR makes it possible to parse older Ruby versions that were accidentally dropped before RuboCop AST was separated from RuboCop core.

With this fix, the following recovery work will be possible.

- Recover for Ruby 2.5 analysis: https://github.com/rubocop/rubocop/pull/10626
- Recover for Ruby 2.4 analysis: https://github.com/rubocop/rubocop/pull/10632

### DIFF
--- a/changelog/new_make_parse_from_ruby_19_to_23_available.md
+++ b/changelog/new_make_parse_from_ruby_19_to_23_available.md
@@ -1,0 +1,1 @@
+* [#233](https://github.com/rubocop-hq/rubocop-ast/pull/233): Make parse from Ruby 1.9 to 2.3 available. ([@koic][])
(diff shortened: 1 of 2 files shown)


marcandre:
Good point, thanks

marcandre:
Release 1.18.0

PR #237, opened 2022-08-05
Fix #macro? for numblock nodes

gsamokovarov:
We can have macros in numblocks. I found this issue while working on broader `numblock` nodes support for cops in rubocop and found out that the`#bare_access_modifier?` method was not giving proper results for modifiers in numblocks. This applies to other methods that depend on the `#macro?` method.

### DIFF
--- a/changelog/new_fix_macro_for_numblock_nodes.md
+++ b/changelog/new_fix_macro_for_numblock_nodes.md
@@ -1,0 +1,1 @@
+* [#237](https://github.com/rubocop/rubocop-ast/pull/237) Fix `#macro?` for numblock nodes ([@gsamokovarov][])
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bug fixes
+
+* [#237](https://github.com/rubocop/rubocop-ast/pull/237) Fix `#macro?` for numblock nodes ([@gsamokovarov][])
```
koic, line 7: Can you use the new changelog entry method instead of editing CHANGELOG.md directly?
https://github.com/rubocop/rubocop/blob/v1.33.0/CONTRIBUTING.md#creating-changelog-entries

gsamokovarov: Surely, Kichi, thanks for noticing. I have run `rake changelog:new` and added the changelog entry for this change.

PR #239, opened 2022-09-16
Add `multiple_elsif` to `IfNode`

koic:
Follow up https://github.com/rubocop/rubocop/pull/11007#discussion_r973140044.

This PR adds `multiple_elsif` to `IfNode`.

### DIFF
--- a/changelog/new_add_multiple_elsif_to_if_node.md
+++ b/changelog/new_add_multiple_elsif_to_if_node.md
@@ -1,0 +1,1 @@
+* [#239](https://github.com/rubocop/rubocop-ast/pull/239): Add `multiple_elsif` method to `AST::IfNode`. ([@koic][])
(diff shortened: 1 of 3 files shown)


marcandre:
Thanks for the PR.
This seems at first sight to be quite specific. I'm wondering if it isn't too specific to of that much used.
For example, do you foresee that will would use this method in many places of `rubocop`'s code?

koic:
I opened this PR because I thought this feature could be made generic. But, I don't have any concrete plans to use outside of https://github.com/rubocop/rubocop/pull/11007#discussion_r973140044.

Do you think it's too early to introduce this in rubocop-ast? If so, I will close the PR because I may not have chosen the timing of the introduction well 😅

marcandre:
Yes, let's do this, we can close this and revisit if when/if this gets used in a few places.

koic:
I get it! Thank you for taking your time!

PR #246, opened 2022-12-06
Optimize `Node#parent_module_name`

fatkodima:
Related - https://github.com/rubocop/rubocop/pull/11240

There is no need to do all the complex traversing if there is no parent.

### DIFF
--- a/lib/rubocop/ast/node.rb
+++ b/lib/rubocop/ast/node.rb
@@ -313,6 +313,8 @@ module RuboCop
       ## Searching the AST
 
       def parent_module_name
+        return 'Object' unless parent
+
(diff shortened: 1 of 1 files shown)


Drenmi (approved):

marcandre:
Thanks for the PR.

I am not convinced this will make any difference, like *ever*.

You would need a _single_ and _top-level_ `def` node for this to be the case, otherwise there will be a parent (`begin`).

I'm closing, but we can reopen if I'm missing something.

fatkodima:
Agreed. Retested - you are right! Probably I was mistaken by something else back then 😄

PR #249, opened 2022-12-13
Several optimizations

fatkodima:
Relevant - https://github.com/rubocop/rubocop/pull/11279.


lib/rubocop/ast/node/mixin/method_dispatch_node.rb
```diff
@@ -211,7 +211,7 @@ def lambda?
       #
       # @return [Boolean] whether this method is a lambda literal
       def lambda_literal?
-        block_literal? && loc.expression && loc.expression.source == '->'
+        loc.expression && loc.expression.source == '->' && block_literal?
```
iliabylich, line 214: are there nodes that don't have `.loc.expression`?

lib/rubocop/ast/node.rb
```diff
@@ -101,11 +105,19 @@ def initialize(type, children = [], properties = {})
...
         method_name = "#{node_type.to_s.gsub(/\W/, '')}_type?"
-        define_method(method_name) do
-          type == node_type
-        end
+        class_eval <<~RUBY, __FILE__, __LINE__ + 1
+          def #{method_name}          # def block_type?
+            @type == :#{node_type}    #   @type == :block
+          end                         # end
+        RUBY
+      end
```
fatkodima, line 115: This code is in the hot path. Executing `class_eval`-defined method is 60% faster than by `define_method`.
Comparing instance variable is 20% faster that calling an `attr_accessor`-defined method.

lib/rubocop/ast/node.rb
```diff
@@ -101,11 +105,19 @@ def initialize(type, children = [], properties = {})
...
+
+      # Most nodes are of 'send' type, so this method is defined
+      # separately to make this check as fast as possible.
+      def send_type?
+        false
       end
```
fatkodima, line 121: Most checks are done against `send` nodes, so I extracted a dedicated fastest method for this.

lib/rubocop/ast/node/mixin/method_dispatch_node.rb
fatkodima: Seems like no. Thank you, removed that check.

marcandre:
Looks great!
I thought I had merged it, sorry for the delay.

marcandre:
Released in 1.14.1

PR #251, opened 2023-01-07
Use concurrency for GitHub Actions workflow

ydah:
Same as: https://github.com/rubocop/rubocop-rspec/pull/1539

This PR ensures that only the last execution per branch always works, even when pushed multiple times in a row, for example. The older one is canceled.

Refs:
https://docs.github.com/en/actions/using-jobs/using-concurrency https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -9,6 +9,10 @@ on:
       - master
   pull_request:
 
+concurrency:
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks!

PR #253, opened 2023-01-09
Improve the code example about `xstr`

ydah:
This PR is improve the code example about `xstr`.

## Motivation

The code example for `xstr` appears to be incorrect from the [documentation](https://docs.rubocop.org/rubocop-ast/node_types.html).

xstr

The `foo#bar` looks more incorrect because it also looks like `dstr` and the backtick is not escaped.
Therefore, we would like to suggest using  \`date\` as a code example, as used in the following https://docs.ruby-lang.org/ document.

https://docs.ruby-lang.org/en/master/Kernel.html#method-i-60:~:text=%24-,%60date%60,-%23%20%3D%3E%20%22Wed%20Apr%20%209

doc

After:
after


ydah:
[memo] need merge https://github.com/rubocop/rubocop-capybara/pull/17

marcandre:
Sorry for the delay.
Good catch 👍

PR #255, opened 2023-02-07
Make `Node#class_constructor?` aware of `Data.define` and numbered parameters

koic:
This PR contains the following two commits:

## Make `Node#class_construcor?` aware of Ruby 3.2's `Data.define`

Ruby 3.2 introduced immutable `Data.define`. Since this is similar to `Struct.new`, `Node#class_constructor` will handle it in the same way.

## Make `Node#class_constructor?` aware of Ruby 2.7's numbered parameters

`Class.new`, `Module.new`, `Struct.new`, and `Data.define` take themselves as a block parameter.
These are the same when using numbered block parameter.

I will try to update https://github.com/rubocop/rubocop/pull/11546 and https://github.com/rubocop/rubocop/pull/11547 with this feature.

### DIFF
--- a/lib/rubocop/ast/node.rb
+++ b/lib/rubocop/ast/node.rb
@@ -513,7 +513,7 @@ module RuboCop
         {
           (send #global_const?({:Class :Module :Struct}) :new ...)
           (send #global_const?(:Data) :define ...)
-          (block {
+          ({block numblock} {
             (send #global_const?({:Class :Module :Struct}) :new ...)
             (send #global_const?(:Data) :define ...)
           } ...)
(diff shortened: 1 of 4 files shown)


Drenmi (approved):
🚀

lib/rubocop/ast/node.rb
```diff
@@ -510,8 +510,14 @@ def guard_clause?
 
       # @!method class_constructor?(node = self)
       def_node_matcher :class_constructor?, <<~PATTERN
-        {       (send #global_const?({:Class :Module :Struct}) :new ...)
-         (block (send #global_const?({:Class :Module :Struct}) :new ...) ...)}
+        {
+          (send #global_const?({:Class :Module :Struct}) :new ...)
+          (send #global_const?(:Data) :define ...)
+          ({block numblock} {
```
marcandre, line 516: This is good, but makes me wonder if we shouldn't have a pseudo type to include both of these... Maybe `any_block` or `a_block` instead of `{block numblock}`

marcandre:
Thanks! 🚀

marcandre:
Released as 1.26

PR #259, opened 2023-03-24
Add `forwarded_kwrestarg` node to `AST::Builder`

koic:
This PR adds `forwarded_kwrestarg` node to `AST::Builder`.

This feature is required to resolve https://github.com/rubocop/rubocop/issues/11730.

I thought about defining a class like `ForwardedKwrestargNode` for `forwarded_kwrestarg`, but the node is actually an anonymous `KeywordSplatNode` so it's used.


lib/rubocop/ast/builder.rb
```diff
@@ -19,76 +19,77 @@ class Builder < Parser::Builders::Default
...
+        class:               ClassNode,
+        const:               ConstNode,
+        def:                 DefNode,
+        defined?:            DefinedNode,
+        defs:                DefNode,
+        dstr:                DstrNode,
+        ensure:              EnsureNode,
+        for:                 ForNode,
+        forward_args:        ForwardArgsNode,
+        forwarded_kwrestarg: KeywordSplatNode,
```
koic, line 55: This line is the only one added in this map.

spec/rubocop/ast/keyword_splat_node_spec.rb
```diff
@@ -39,6 +39,82 @@
...
+    let(:kwsplat_node) { parse_source(source).ast.children.last.children.last }
+
+    describe '.new' do
+      let(:source) do
+        <<~RUBY
+          def foo(**)
+            { a: 1, ** }
+          end
+        RUBY
+      end
```
marcandre, line 52: Could we share the `let(:source)`? It would also be closer to the `let(:kwsplat_node)`

koic: Ah, makes sense! I've updated it. Thank you for the suggestion!

marcandre:
My only hesitation is: do we want a different class, or could we not use `Kwsplat`?

koic:
Ah, yeah, I thought it would be better to be able to distinguish between `kwsplat` node type and `forwarded_kwrestarg` node type, but it seems pragmatic and more better to have compatibility with `kwsplat_type?` as you suggested. I've updated this PR.

spec/rubocop/ast/keyword_splat_node_spec.rb
```diff
@@ -39,6 +39,45 @@
...
+      it { expect(kwsplat_node.value).to eq(kwsplat_node) }
+    end
+
+    describe '#operator' do
+      it { expect(kwsplat_node.operator).to eq('**') }
+    end
+
+    describe '#kwsplat_type?' do
+      it { expect(kwsplat_node).to be_kwsplat_type }
+    end
```
marcandre, line 78: Great.
We also need `it { expect(kwsplat_node).to be_forwarded_kwrestarg_type }`, right?

koic: I see! surely! I've add specs for that.

>  distinguish between `kwsplat` node type and `forwarded_kwrestarg` node type

This worry of mine was unfounded 😅

marcandre:
Thank you so much! Released as 1.28.0 🚀

PR #267, opened 2023-05-30
Introduce RuboCop::Ast::MethodDispatchNode#selector

gsamokovarov:
If you need to manipulate the call-site of a method dispatch, you have to use `node.loc.selector` for regular methods and `node.loc.keyword` for syntax nodes that behave like method invocations. Super and yield nodes are nodes are examples.

This forces us to check if `node.loc` responds to `#keyword` in a few places in rubocop proper and we can make it easier if we expose a single method called `#selector` (the name is up for debate) and use it for method dispatch nodes.

As an example, see [HashSyntaxShorthand](https://github.com/rubocop/rubocop/blob/7259df96e689f80828eb231495cc515011585bd9/lib/rubocop/cop/mixin/hash_shorthand_syntax.rb#L206-L224) where I had to define a struct to hide away the selector detection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#262](https://github.com/rubocop/rubocop-ast/pull/267): Introduce RuboCop::Ast::MethodDispatchNode#selector. ([@gsamokovarov][])
+
 ## 1.28.1 (2023-05-01)
 
 ### Bug fixes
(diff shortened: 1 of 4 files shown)


marcandre:
Thanks for the PR @gsamokovarov .

No problem for `rubocop-ast`, but for contributions for `rubocop`, please check [the preferred way](https://docs.rubocop.org/rubocop/about/changelog.html) to make entries.

I think this is a good idea, but would I would like to know what @koic thinks about it

.rubocop.yml
```diff
@@ -97,6 +97,7 @@ Metrics/BlockLength:
     - '**/*.gemspec'
 
 Metrics/ModuleLength:
+  Max: 114
```
bbatsov, line 100: That looks a bit random and kind of unrelated to the rest of the commit. :-)

bbatsov:
For the record - it looks like a reasonable addition to me. Frankly, I don't see much point in `parser` treating those differently.

.rubocop.yml
gsamokovarov: Fair enough, it's the length of the current `MethodDispatchNode` module. Should I disable the cop for the module and revert the config?

gsamokovarov:
Thank you! ❤️

PR #272, opened 2023-11-09
Make `Node#left_curly_brace?` aware of lambda brace

koic:
Follow up https://github.com/rubocop/rubocop/pull/12376.

This PR makes `Node#left_curly_brace?` aware of lambda brace (`-> {f}`).


bbatsov:
I think that's a good change. Thanks!

PR #276, opened 2024-01-15
Support `Parser::Ruby34`

koic:
Parser gem has been started development for Ruby 3.4 (edge Ruby).
https://github.com/whitequark/parser/pull/991

And this PR update to require Parser 3.3.0.4 or higher, which contains `Parser::Ruby34`. https://github.com/whitequark/parser/blob/master/CHANGELOG.md#not-released-2024-01-15


bbatsov:
Thanks!

PR #277, opened 2024-02-24
Support Prism as a Ruby parser

koic:
This PR introduces the `parser_engine` option to `ProcessedSource` to support Prism, as part of the RuboCop AST side effort towards addressing https://github.com/rubocop/rubocop/issues/12600.

## Configuration

By default, analysis is performed using the Parser gem, so the default value for the newly added `parser_engine` is `parser_whitequark`:

```ruby
ProcessedSource.new(@options[:stdin], ruby_version, file, parser_engine: :parser_whitequark)
```

This code maintains compatibility, meaning the traditional behavior is preserved:

```ruby
ProcessedSource.new(@options[:stdin], ruby_version, file)
```

To perform analysis using Prism, specify `parser_engine: :parser_prism`:

```ruby
ProcessedSource.new(@options[:stdin], ruby_version, file, parser_engine: :parser_prism)
```

The parameter name `parser_prism` reflects the original parser_prism which was the basis for `Prism::Translation::Parser` (now integrated into Prism): https://github.com/kddnewton/parser-prism

This is an experimental introduction, and some incompatibilities still remain.

> [!NOTE]
> As initially mentioned in https://github.com/rubocop/rubocop/issues/12600#issuecomment-1933657732,
> the plan was to set `parser_engine: prism`.
>
> However, the parser engine used in this PR is `Prism::Translation::Parser`, not `Prism`:
> https://github.com/ruby/prism/pull/2419
>
> `Prism::Translation::Parser` and `Prism` have different ASTs, so their migration will definitely cause incompatibility.
> So, considering the possibility of further replacing `Prism::Translation::Parser` with `Prism` in the future,
> it has been decided that it might be better not to use `ParserEngine: prism` for the time being.
> `ParserEngine: prism` is reserved for `Prism`, not `Prism::Translation::Parser`.
>
> Therefore, the parameter value has been set to `parser_engine: parser_prism` specifically for
> `Prism::Translation::Parser`.
>
> This means that the planned way to specify Prism in .rubocop.yml file will be `ParserEngine: parser_prism`,
> not `ParserEngine: prism`.

## Compatibility

The compatibility issues between Prism and the Parser gem have not been resolved. The failing tests will be skipped with `broken_on: :prism`:

- https://github.com/ruby/prism/issues/2454 has been resolved but not yet released.
- https://github.com/ruby/prism/issues/2467 is still unresolved.

Issues that will be resolved in several upcoming releases of Prism are being skipped with `broken_on: :prism`.

Anyway, RuboCop AST can be released independently of the resolution and release of Prism.

> [!NOTE]
> The hack in `Prism::Translation::Parser` for `ProcessedSource` needs to be fixed:
> https://github.com/ruby/prism/blob/v0.24.0/lib/prism/translation/parser/rubocop.rb
>
> If the above interface is accepted, a fix will be proposed on the Prism side.

## Test

Tests for RuboCop AST with Prism as the backend can be run as follows:

```console
bundle exec rake prism_spec
```

The above is the shortcut alias for:

```console
PARSER_ENGINE=parser_prism TARGET_RUBY_VERSION=3.3 rake spec
```

RuboCop AST works on Ruby versions 2.6+, but since Prism only targets analysis for Ruby 3.3+, `internal_investigation` Rake task will not be executed. This task is only run with the Parser gem, which can analyze Ruby versions 2.0+.

### DIFF
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -15,15 +15,23 @@ module RuboCop
       INVALID_LEVELS = %i[error fatal].freeze
       private_constant :INVALID_LEVELS
 
-      attr_reader :path, :buffer, :ast, :comments, :tokens, :diagnostics,
-                  :parser_error, :raw_source, :ruby_version
+      PARSER_ENGINES = %i[parser_whitequark parser_prism].freeze
+      private_constant :PARSER_ENGINES
 
-      def self.from_file(path, ruby_version)
+      attr_reader :path, :buffer, :ast, :comments, :tokens, :diagnostics,
+                  :parser_error, :raw_source, :ruby_version, :parser_engine
+
+      def self.from_file(path, ruby_version, parser_engine: :parser_whitequark)
         file = File.read(path, mode: 'rb')
-        new(file, ruby_version, path)
+        new(file, ruby_version, path, parser_engine: parser_engine)
       end
 
-      def initialize(source, ruby_version, path = nil)
+      def initialize(source, ruby_version, path = nil, parser_engine: :parser_whitequark)
+        unless PARSER_ENGINES.include?(parser_engine)
+          raise ArgumentError, 'The keyword argument `parser_engine` accepts ' \
+                               "`parser` or `parser_prism`, but `#{parser_engine}` was passed."
+        end
+
         # Defaults source encoding to UTF-8, regardless of the encoding it has
         # been read with, which could be non-utf8 depending on the default
         # external encoding.
@@ -33,9 +41,10 @@ module RuboCop
         @path = path
         @diagnostics = []
         @ruby_version = ruby_version
+        @parser_engine = parser_engine
         @parser_error = nil
 
-        parse(source, ruby_version)
+        parse(source, ruby_version, parser_engine)
       end
 
       def ast_with_comments
@@ -193,7 +202,7 @@ module RuboCop
         end
       end
 
-      def parse(source, ruby_version)
+      def parse(source, ruby_version, parser_engine)
         buffer_name = @path || STRING_SOURCE_NAME
         @buffer = Parser::Source::Buffer.new(buffer_name, 1)
 
@@ -207,7 +216,7 @@ module RuboCop
           return
         end
 
-        @ast, @comments, @tokens = tokenize(create_parser(ruby_version))
+        @ast, @comments, @tokens = tokenize(create_parser(ruby_version, parser_engine))
       end
 
       def tokenize(parser)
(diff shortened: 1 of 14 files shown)


koic:
Prism requires Ruby 2.7 or higher at runtime. The most straightforward solution would be to drop Ruby 2.6 from RuboCop AST's runtime support. I plan to open a PR to drop Ruby 2.6 from the runtime. Do you have any thoughts on this?

bbatsov:
That's fine by me. Ruby 2.6's usage is quite low already and the upgrade that to 2.7 is trivial.

lib/rubocop/ast/processed_source.rb
```diff
@@ -227,61 +236,77 @@ def tokenize(parser)
...
-          Parser::Ruby33
-        when 3.4
-          require 'parser/ruby34'
-          Parser::Ruby34
-        else
-          raise ArgumentError,
-                "RuboCop found unknown Ruby version: #{ruby_version.inspect}"
+      def parser_class(ruby_version, parser_engine)
+        case parser_engine
+        when :parser
```
bbatsov, line 241: I wonder if we shouldn't name this `:parser_whitequark`, give that `parser` is a very generic name.

bbatsov:
I like the PR overall - no concerns from me regarding the overall approach and the implementation. @rubocop/rubocop-core please take a look as well, given this is a major development. 

@koic Did you get the chance to compare the performance with the new old and the new parser engine?

lib/rubocop/ast/processed_source.rb
koic: `parser_whitequark` is a unique name, but the context might become distant if the org of `whitequark/parser` changes to something other than whitequark.

Maybe something like `parser_gem` can be suggested, but I'm not that confident about that. So, `parser_whitequark` is fine too, if @bbatsov prefers that :-)

koic:
> Did you get the chance to compare the performance with the new old and the new parser engine?

Here is an example of benchmark for my local development:

```ruby
# bench.rb
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'benchmark-ips'
  gem 'rubocop-ast', path: '/Users/koic/src/github.com/rubocop/rubocop-ast'
end

require 'rubocop-ast'

source = File.read(__FILE__)

Benchmark.ips do |x|
  x.report('Parser') do
    RuboCop::AST::ProcessedSource.new(source, 3.3, parser_engine: :parser)
  end

  x.report('Prism') do
    RuboCop::AST::ProcessedSource.new(source, 3.3, parser_engine: :parser_prism)
  end

  x.compare!
end
```

Although results may depending on the code analyzed, it has been observed that Prism tends to be faster:

```console
$ ruby bench.rb
Fetching gem metadata from https://rubygems.org/..
Resolving dependencies...
ruby 3.4.0dev (2024-02-24T04:19:37Z master e9e752c7ef) [x86_64-darwin23]
Warming up --------------------------------------
              Parser    62.000 i/100ms
               Prism   107.000 i/100ms
Calculating -------------------------------------
              Parser    607.475 (± 4.9%) i/s -      3.038k in   5.013690s
               Prism      1.040k (± 3.0%) i/s -      5.243k in   5.047621s

Comparison:
               Prism:     1039.7 i/s
              Parser:      607.5 i/s - 1.71x  slower
```

lib/rubocop/ast/processed_source.rb
koic: For now, I've renamed it to `parser_whitequark`. Please let me know if you'd like the name updated further.

bbatsov: I think that name is fine and we can always provide some aliases down the road. 

In general, I'm guessing that at some point native Prism (without the `parser_whitequark`) translation will be the only parsing engine we'll be supporting. I guess that once most tools switch to Prism maintaining the older parsers will become pointless. It's amazing how quickly Prism is getting traction in the Ruby community these days https://blog.jruby.org/2024/02/23/jruby-prism-parser.html

bbatsov:
One step closer! Great work, @koic! 🔥

marcandre:
Looks very to me 🚀

koic:
@bbatsov @marcandre Can you release this feature to move forward with https://github.com/rubocop/rubocop/pull/12724?

bbatsov:
Sure, I'll do this right away.

PR #279, opened 2024-02-24
Drop Ruby 2.6 runtime support

koic:
Follow up https://github.com/rubocop/rubocop-ast/pull/277#issuecomment-1962302045.

This PR drops Ruby 2.6 runtime support to prepare to support Prism.


bbatsov:
Looks good to me!

PR #283, opened 2024-03-01
Remove prism as a runtime dependency

### LINKED ISSUE #282: Experimental Prism feature listed as runtime dependency
Support for Prism was added in rubocop-ast 1.31.0 and marked as "experimental" in the changelog. However, Prism was also added as a runtime dependency in the Gemspec.

It seems contradictory to require a library for an experimental feature, particularly when it's only available in the latest Ruby release.

We're starting to see some failures in our CI pipelines because of this new requirement.

Would you be able to remove Prism as a runtime dependency? Thank you!

marcandre:
Fixes #282

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ gem 'rspec', '~> 3.7'
 # Stop upgrading SimpleCov until the following issue will be resolved.
 # https://github.com/codeclimate/test-reporter/issues/418
 gem 'simplecov', '~> 0.10', '< 0.18'
+gem 'prism', '>= 0.24.0', require: false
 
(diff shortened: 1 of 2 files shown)


koic:
It is mentioned below, it seems better to hold off on making a decision to remove it for now:
https://github.com/rubocop/rubocop-ast/issues/282#issuecomment-1972341899

If the decision is made to remove it, it would likely require adjustments not only to the removing dependency, also to issues surrounding Prism's loading and documentation. It might be better if I handle it. Thank you.

koic:
I've opened #285. It addresses users who do not have Prism installed, indicating through warning and documentation that Prism needs to be installed separately.

bbatsov:
Addressed by #285.

PR #285, opened 2024-03-01
[Fix #282] Remove Prism from runtime dependency

### LINKED ISSUE #282: Experimental Prism feature listed as runtime dependency

koic:
Fixes #282.

This PR removes Prism from runtime dependency.

If it is decided that Prism will not be a runtime dependency for the time being, error message and documentation will be used to communicate the dependency on Prism to users.

Making it a default runtime dependency will be avoided until at least #282 installation error with Prism is resolved.

### DIFF
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -286,7 +286,12 @@ module RuboCop
             raise ArgumentError, "RuboCop found unknown Ruby version: #{ruby_version.inspect}"
           end
         when :parser_prism
-          require 'prism'
+          begin
+            require 'prism'
+          rescue LoadError
+            warn "Error: Unable to load Prism. Add `gem 'prism'` to your Gemfile."
+            exit!
+          end
 
           case ruby_version
           when 3.3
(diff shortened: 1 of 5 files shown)


lib/rubocop/ast/processed_source.rb
```diff
@@ -286,7 +286,12 @@ def parser_class(ruby_version, parser_engine)
             raise ArgumentError, "RuboCop found unknown Ruby version: #{ruby_version.inspect}"
           end
         when :parser_prism
-          require 'prism'
+          begin
+            require 'prism'
+          rescue LoadError
+            warn "Error: Unable to load Prism. Add `gem 'prism'` to your Gemfile."
```
sambostock, line 292: Do we need to enforce a minimum version?

PR #286, opened 2024-03-08
Improve error message for invalid `parser_engine` value

Earlopain:
`parser` doesn't exist, it must be `parser_whitequark`. In addition, this only accepts symbols.

When passing a string, this would previously print ``The keyword argument `parser_engine` accepts `parser` or `parser_prism`, but `parser_prism` was passed.``. Use `inspect` to clearly show what you did wrong.

The new error message looks like this: ``The keyword argument `parser_engine` accepts `:parser_whitequark` or `:parser_prism`, but `"parser_prism"` was passed.``.

### DIFF
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -27,9 +27,10 @@ module RuboCop
       end
 
       def initialize(source, ruby_version, path = nil, parser_engine: :parser_whitequark)
+        parser_engine = parser_engine.to_sym
         unless PARSER_ENGINES.include?(parser_engine)
-          raise ArgumentError, 'The keyword argument `parser_engine` accepts ' \
-                               "`parser` or `parser_prism`, but `#{parser_engine}` was passed."
+          raise ArgumentError, 'The keyword argument `parser_engine` accepts `parser_whitequark` ' \
+                               "or `parser_prism`, but `#{parser_engine}` was passed."
         end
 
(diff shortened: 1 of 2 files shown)


koic:
It seems to be issues with `Layout/LineLength`.
https://github.com/rubocop/rubocop-ast/actions/runs/8201827148/job/22431355671?pr=286

Earlopain:
Yes, sorry. I already force-pushed.

bbatsov:
I wonder if we shouldn't convert strings to keywords automatically here (if passed), as it's not a big deal.

Earlopain:
That makes sense to me. Reverted most of the error message changes since they are now unnecessary and made it to instead accept strings as well.

lib/rubocop/ast/processed_source.rb
```diff
@@ -27,9 +27,10 @@ def self.from_file(path, ruby_version, parser_engine: :parser_whitequark)
       end
 
       def initialize(source, ruby_version, path = nil, parser_engine: :parser_whitequark)
+        parser_engine = parser_engine.to_sym if parser_engine.is_a?(String)
```
bbatsov, line 30: I recall `.to_sym` is defined on symbols, so you don't really need the `if` here.

marcandre:
Released in v1.31.2, thanks!

koic:
I had initially decided to accept only symbol argument, designing it would lead to better performance through consistency. But, currently, I think the convenience of accepting both string and symbol argument might be more beneficial. Let's accept both 👍

PR #288, opened 2024-03-14
Fix a ruby warning when executing with `bundle exec rubocop` in this repository

Earlopain:
`check_commit.rake` requires 'English' but when doing `bundle exec rubocop` this repository this file will also be evaluated.

This also happens in the `rubocop` repository (my main motivation here) if `rubocop-ast` is in the parent folder because of this line in its Gemfile: https://github.com/rubocop/rubocop/blob/66c8276b5d9e60fed1fb8d937649c3c153e5037d/Gemfile#L29-L30

```
$ RUBYOPT=-w bundle exec rubocop -V
.../rubocop-ast/rubocop-ast.gemspec:18: warning: global variable `$RS' not initialized
1.62.1 (using Parser 3.3.0.5, rubocop-ast 1.31.2, running on ruby 3.3.0) [x86_64-linux]
  - rubocop-performance 1.20.2
  - rubocop-rspec 2.27.1
```

### DIFF
--- a/rubocop-ast.gemspec
+++ b/rubocop-ast.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
 
   s.email = 'rubocop@googlegroups.com'
   s.files = `git ls-files lib LICENSE.txt README.md`
-            .split($RS) + %w[
+            .lines(chomp: true) + %w[
               lib/rubocop/ast/node_pattern/parser.racc.rb
               lib/rubocop/ast/node_pattern/lexer.rex.rb
             ]


Earlopain:
Yeah, sure. Updated

bbatsov:
Can you update this PR so I can finally have it merged?

Earlopain:
Yeah, sure. I changed the solution, can't actually use `$/` since there is a cop for discouraging these constants. Instead I use `lines(chomp: true)` which results in the same thing, `$/` is the default separator.

bbatsov:
Thanks for addressing this so quickly! 🙇

PR #295, opened 2024-06-19
Move test `Node#used?` predicate method definition

sambostock:
Definining a method in a `before :all` block doesn't clean it up automatically, meaning once we've defined it, it's there for all remaining tests.

Since it leaks anyway, we can just define it at the top of the file, and avoid needing to disable `ConstantDefinitionInBlock`.

### DIFF
--- a/spec/rubocop/ast/node_spec.rb
+++ b/spec/rubocop/ast/node_spec.rb
@@ -2,25 +2,23 @@
 
 require 'uri'
 
+module RuboCop
+  module AST
+    # Patch Node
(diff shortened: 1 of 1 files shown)


marcandre:
I have no idea why CI is not green though...

koic:
> I have no idea why CI is not green though...

The CI error has been resolved in https://github.com/rubocop/rubocop-ast/commit/56ba9fe4c30aef30ea74ab47375a958f926c9283, so please rebase with the latest master branch.

PR #298, opened 2024-06-20
Define `recursive_*literal?` methods using macro

sambostock:
This allows us to document the methods, without needing to repeat the implementations.

This is the same approach as #297.


marcandre:
Very nice, thank you.

PR #299, opened 2024-06-20
Move `rubocop:disable` comments out of documentation

sambostock:
When `rubocop:disable` is used immediately above a method or module declaration, it becomes part of the documentation.

e.g. [`Node#value_used?`](https://rubydoc.info/github/rubocop/rubocop-ast/RuboCop%2FAST%2FNode:value_used%3F)

image

Instead, if we inline it with the declaration itself, it is not included in the documentation. As a side benefit, RuboCop attaches it to that declaration only, removing the need to `rubocop:enable` afterwards.

However, YARD applies it as documentation to whatever immediately follows, so if the next declaration is public, we must explicitly document it, so YARD properly ignores the `rubocop:disable` comment.


lib/rubocop/ast/node_pattern/compiler/sequence_subcompiler.rb
```diff
@@ -13,8 +13,8 @@ class Compiler
         # Doc on how this fits in the compiling process:
         #   /docs/modules/ROOT/pages/node_pattern.adoc
         #
-        # rubocop:disable Metrics/ClassLength
-        class SequenceSubcompiler < Subcompiler
+        class SequenceSubcompiler < Subcompiler # rubocop:disable Metrics/ClassLength
+          # Shift of 1 from standard Ruby indices
```
sambostock, line 17: Inferred from a comment later in this file.

lib/rubocop/ast/node_pattern/compiler/binding.rb
```diff
@@ -26,8 +26,9 @@ def bind(name)
             var
           end
 
-          # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
-          def union_bind(enum)
+          # Yields for each branch of the given union, forbidding unification of
+          # bindings which only appear in a subset of the union.
+          def union_bind(enum) # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
```
sambostock, line 31: I'm not familiar with this section of the code base and what this method is actually meant to do, so I'm not sure this comment is accurate.

marcandre:
Thank you.

PR #302, opened 2024-07-10
[Docs] Update the doc of Node Types

koic:
Follow up https://github.com/whitequark/parser/pull/716.

The `def_e` and `defs_e` nodes representing endless method definitions no longer exist. Instead, these are represented by `def` and `defs` nodes with empty `location.end` information.


marcandre:
Thank you 👍

PR #303, opened 2024-07-10
Remove non-existent nodes from document

yui-knk:
In parser gem endless method definition is mapped to `def` node or `defs` node. Then there is no `def_e` node nor `defs_e` node.

These nodes types were once introduced by https://github.com/whitequark/parser/pull/676 however emitted by https://github.com/whitequark/parser/pull/716.


yui-knk:
Oh, this is duplicated with https://github.com/rubocop/rubocop-ast/pull/302 ...

PR #307, opened 2024-08-03
Remove `Range#minmax` refinement

Earlopain:
It was added for a bug in Ruby < 2.7
The minimum version now is 2.7 so it is not needed anymore.

Context in https://github.com/rubocop/rubocop-ast/pull/151 and https://github.com/rubocop/rubocop-ast/pull/150

```
irb(main):001:0> RUBY_VERSION
=> "2.6.10"
irb(main):002:0> Range.instance_method(:minmax).owner != ::Range
=> true

irb(main):002> RUBY_VERSION
=> "2.7.8"
irb(main):003> Range.instance_method(:minmax).owner != ::Range
=> false
```


koic:
Thanks!

PR #308, opened 2024-08-08
Fix readme CI badge

Earlopain:
Not quite sure where this currently is getting its data from but its not correct.

Also, the codeclimate badge links in the readme are broken. This is probably because of the `rubocop-hq` move. See https://github.com/rubocop/rubocop/pull/12946#issuecomment-2139390046


marcandre:
Thank you ♥️

PR #309, opened 2024-08-17
Mark `RuboCop::AST::EnsureNode` as being in a void context.

Earlopain:
`ensure` nodes are somewhat special in that their last expression will not get implicitly returned.

```rb
def foo(do_rescue:)
  puts "begin called"
  raise if do_rescue
  "begin"
rescue
  puts "rescue called"
  "rescue"
ensure
  puts "ensure called"
  "ensure"
end

puts foo(do_rescue: false)
puts "----"
puts foo(do_rescue: true)

# =>
# begin called
# ensure called
# begin
# ----
# begin called
# rescue called
# ensure called
# rescue
```

This change will make it possible to properly teach `Lint/Void` about it.

### DIFF
--- a/changelog/change_mark_ensure_as_void_context.md
+++ b/changelog/change_mark_ensure_as_void_context.md
@@ -1,0 +1,1 @@
+* [#309](https://github.com/rubocop/rubocop-ast/pull/309): Mark `RuboCop::AST::EnsureNode` as being in a void context. ([@earlopain][])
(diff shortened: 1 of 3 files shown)


marcandre:
Good catch, thanks!

marcandre:
Released as 1.32.1

Earlopain:
Thank you!

PR #310, opened 2024-08-17
Fix `RuboCop::AST::DefNode#void_context?` to handle class methods called `initialize`

vlad-pisanov:
Currently, `DefNode#void_context?` incorrectly returns `true` for class methods called `initialize`. This PR fixes this.

```ruby
# constructor: void context ✔️ 
def initialize
  42
end

# class method: not void context! ❌ 
def self.initialize
  42
end
```

### DIFF
--- a/lib/rubocop/ast/node/def_node.rb
+++ b/lib/rubocop/ast/node/def_node.rb
@@ -13,7 +13,7 @@ module RuboCop
       #
       # @return [Boolean] whether the `def` node body is a void context
       def void_context?
-        method?(:initialize) || assignment_method?
+        (def_type? && method?(:initialize)) || assignment_method?
       end
 
       # Checks whether this method definition node forwards its arguments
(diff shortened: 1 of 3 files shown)


lib/rubocop/ast/node/def_node.rb
```diff
@@ -13,7 +13,7 @@ class DefNode < Node
       #
       # @return [Boolean] whether the `def` node body is a void context
       def void_context?
-        method?(:initialize) || assignment_method?
+        (def_type? && method?(:initialize)) || assignment_method?
```
Earlopain, line 16: Conceptually this makes sense to me, though I wonder if would be better to instead introduce a `DefsNode` instead which then overrides this method

vlad-pisanov: I actually thought `DefsNode` was already defined going into this.

I'd rather not make drastic changes in this PR. This is a pre-existing pattern (used in `BlockNode` for example):

https://github.com/rubocop/rubocop-ast/blob/30439dbd930ca7a68e83b83fe884226a8747aad8/lib/rubocop/ast/node/block_node.rb#L49

Earlopain: Ah, right. Yeah, probably fine then. I did a quick search around but didn't find other nodes that did it like this but clearly I didn't search good enough.

marcandre:
Sorry for the delay, I missed the notification.

Thanks for the PR 👍

My cell phone just died on me, so I can't release right now, but will in a few days.

marcandre:
Sure, that'd be great, thanks @koic ❤️

PR #313, opened 2024-09-01
Optimize forwarded calls

marcandre:
Fixes [#312]

@Earlopain do you have a minute to review the PR? 🙏

### DIFF
--- a/lib/rubocop/ast.rb
+++ b/lib/rubocop/ast.rb
@@ -5,6 +5,7 @@ require 'forwardable'
 require 'set'
 
 require_relative 'ast/ext/range'
+require_relative 'ast/utilities/simple_forwardable'
 require_relative 'ast/node_pattern/method_definer'
 require_relative 'ast/node_pattern'
 require_relative 'ast/node/mixin/descendence'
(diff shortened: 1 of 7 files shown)


bbatsov:
Did you manage to get the same 3% speedup?

Earlopain:
This looks fine to me, benchmarking confirms:

```
Current:
$ hyperfine -w 3 -r 20 "bundle exec rubocop Rakefile" 
Benchmark 1: bundle exec rubocop Rakefile
  Time (mean ± σ):      2.907 s ±  0.133 s    [User: 2.458 s, System: 0.242 s]
  Range (min … max):    2.802 s …  3.408 s    20 runs

PR:  
$ hyperfine -w 3 -r 20 "bundle exec rubocop Rakefile" 
Benchmark 1: bundle exec rubocop Rakefile
  Time (mean ± σ):      2.695 s ±  0.084 s    [User: 2.297 s, System: 0.229 s]
  Range (min … max):    2.596 s …  2.918 s    20 runs
```

This is with the changes from rubocop itself already applied. It's a weaker machine than the other benchmark so numbers are different and apparently matter even more since it's about 7% just with this change. I haven't tested but it's probably that file system operations are similarly fast while the CPU is not (or something similar to that effect).

Either way, still achives the original goal. Thanks!

marcandre:
Thanks!

Released as `1.32.2`

I'd recommend using that in `rubocop` too.

Issue #319, opened 2024-10-07
Some predicates return `nil`


marcandre:
Found by @dvandersluis in #317.

Unfortunately our specs [don't catch these](https://github.com/rubocop/rubocop-rspec/issues/1966).

My plan:
1) Turn on strict_predictes rspec config
2) Add `def_node_predicate` to be used instead of `def_node_matcher` where it makes sense
3) Fix spec failures
4) analyse the remaining cases which are either redundant, or missing a test.

```
Offenses:

  lib/rubocop/ast/node/array_node.rb:37:9: C: [Correctable] Style/ReturnNilInPredicateMethodDefinition: Predicate method may return nil due to &. operator.
  loc.begin&.is?('[')
^^^^^^^^^^^^^^^^^^^
  lib/rubocop/ast/node/array_node.rb:55:11: C: Style/ReturnNilInPredicateMethodDefinition: Predicate method may return nil due to &. operator.
  loc.begin&.source&.start_with?('%')
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  lib/rubocop/ast/node/block_node.rb:93:9: C: [Correctable] Style/ReturnNilInPredicateMethodDefinition: Predicate method may return nil due to &. operator.
  loc.end&.is?('}')
^^^^^^^^^^^^^^^^^
  lib/rubocop/ast/node/block_node.rb:100:9: C: [Correctable] Style/ReturnNilInPredicateMethodDefinition: Predicate method may return nil due to &. operator.
  loc.end&.is?('end')
^^^^^^^^^^^^^^^^^^^
  lib/rubocop/ast/node/for_node.rb:20:9: C: [Correctable] Style/ReturnNilInPredicateMethodDefinition: Predicate method may return nil due to &. operator.
  loc.begin&.is?('do')
^^^^^^^^^^^^^^^^^^^^
  lib/rubocop/ast/node/hash_node.rb:118:9: C: [Correctable] Style/ReturnNilInPredicateMethodDefinition: Predicate method may return nil due to &. operator.
  loc.end&.is?('}')
^^^^^^^^^^^^^^^^^
  lib/rubocop/ast/node/in_pattern_node.rb:27:9: C: [Correctable] Style/ReturnNilInPredicateMethodDefinition: Predicate method may return nil due to &. operator.
  loc.begin&.is?('then')
^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #320, opened 2024-10-16
Document named captures

Earlopain:
It's a nifty feature currently not mentioned in the docs. Hash keys can be interpolated which would then not make this match anymore but for demonstration purposes this is the best example I could come up with.

cc @dvandersluis 

Unrelated to this, I've added ruby code styling to a code block that was previously missing it.

### DIFF
--- a/docs/modules/ROOT/pages/node_pattern.adoc
+++ b/docs/modules/ROOT/pages/node_pattern.adoc
@@ -126,6 +126,7 @@ Ruby code with a method call with two integer literals as arguments, `foo(1, 2)`
 To match just those method calls where the first argument is a literal `1`, use `(send nil? :foo (int 1) int)`.
 Any child that is a node can be a target for nested matching.
 
+[#any-single-node]
 == `_` for any single node
 
 `_` will check if there's something present in the specific position, no matter the
@@ -134,6 +135,22 @@ value:
 * `(int _)` will match any number
 * `(int _ _)` will not match because `int` types have just one child that
 contains the value.
+
+You can specify a name to make a more descriptive reference:
+
+----
+(send nil? _method_name)
+----
+
+You can also reference them later in the pattern to match against the value that was previously captured:
+
+----
+(pair
+  (_ _key)
+  (_ _key))
+----
+
+`{ a: :a }` will match, while `{ a: :b }` won't.
 
 == `+...+` for several subsequent nodes
 
@@ -314,6 +331,17 @@ The following pattern will have two captures, both arrays:
 (send nil? $int+ (send $...))
 ----
 
+When capturing <<any-single-node, any single node>>, you can reference the value you previously captured.
+
+The following pattern will have one capture:
+
+----
+(pair
+  (_ $_key)
+  (_ _key))
+----
+
+
 == `^` for parent
 
 One may use the `^` character to check against a parent.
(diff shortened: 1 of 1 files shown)


Earlopain:
> without capturing _key

Oh, interesting. It's probably better to move it to https://docs.rubocop.org/rubocop-ast/node_pattern.html#_-for-any-single-node then. If the `..._name` pattern is intentionally supported it'll not be far of from where it was introduced.

docs/modules/ROOT/pages/node_pattern.adoc
```diff
@@ -135,6 +136,22 @@ value:
 * `(int _ _)` will not match because `int` types have just one child that
 contains the value.
 
+You can specify a name name to make them more descriptive:
+
+----
+(send nil? $_method_name)
```
dvandersluis, line 142: Probably meant to not capture in the example here

```suggestion
(send nil? _method_name)
```

dvandersluis (approved):
Looks really good to me! Thanks @Earlopain

docs/modules/ROOT/pages/node_pattern.adoc
Earlopain: Oops, yes. A bit early in the doc for making references to that

marcandre:
> Something to point out is that you can do this without actually making them captures, ie
> 
> ```
> (pair
>   (_ _key)
>   (_ _key))
> ```
> 
> will also work to ensure the same value in both spots, without capturing `_key`.

Correct. Thanks for the PR 🚀

> Also (as discussed on the discord), this also works: `..._foo` (multiple nodes as a named variable which can be referenced -- captures don't seem to work properly though!) . @marcandre this isn't documented or tested and I'm not sure if it's purposeful?

That's incorrect. The lexer is often not space sensitive. In this case, `..._foo` is actually treated the same as `...    _foo`, so `_foo` refers to a single last element (which must exist for a match), not to the variable number of remaining elements (which could be 0).

Would it be helpful to support named "rest"? I think it should look like `...name` and `$...name`. I'd have to modify a bit the lexer and the parser, and maybe compiler to, not sure. If so, we can create a separate issue for the feature request; I'll close this one.

marcandre:
Note: I'm not too sure how to force the update of the docs, but I'm planning on resolving #319 soonish and will release that.

dvandersluis:
> That's incorrect. The lexer is often not space sensitive. In this case, ..._foo is actually treated the same as ...    _foo, so _foo refers to a single last element (which must exist for a match), not to the variable number of remaining elements (which could be 0).
>
> Would it be helpful to support named "rest"? I think it should look like ...name and $...name. I'd have to modify a bit the lexer and the parser, and maybe compiler to, not sure. If so, we can create a separate issue for the feature request; I'll close this one.

Got it, that makes sense. I was pointing it out as an observation I had when playing with the pattern tester, but I didn't look into it closer. Thanks for the explanation. 

I think it'd make sense in the abstract to support `...name` to go with `_name` but I don't know if there's a specific requirement for it.

marcandre:
> I think it'd make sense in the abstract to support `...name` to go with `_name` but I don't know if there's a specific requirement for it.

Ok cool. Maybe we could mention in the docs that it's not supported but to open an issue if there's ever an actual use case?

PR #325, opened 2024-10-31
Allow `non_bare_access_modifier_declaration?` to handle modifiers with multiple arguments

dvandersluis:
I noticed this when testing `Style/AccessModifierDeclarations`. When an access modifier has multiple arguments, `non_bare_access_modifier_declaration?` erroneously (IMO) returns false, which causes the cop to not recognize all cases it should.

The cop does not properly recognize (as a positive or negative, depends on the `AllowModifiersOnSymbols` configuration) this despite the examples for the cop including it.

```ruby
private :foo, :bar
```

I have a follow up PR to `Style/AccessModifierDeclarations` coming once this can be merged.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -92,3 +92,11 @@ RSpec/MultipleExpectations:
 # URISchemes: http, https
 Layout/LineLength:
   Max: 102
+
+# Offense count: 1
+# This cop supports unsafe autocorrection (--autocorrect-all).
+# Configuration parameters: EnforcedStyle, AllowModifiersOnSymbols, AllowModifiersOnAttrs.
+# SupportedStyles: inline, group
+Style/AccessModifierDeclarations:
+  Exclude:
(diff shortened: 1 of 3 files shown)


dvandersluis:
Hmm there's a `Style/AccessModifierDeclarations` failure in `internal_investigation` now because of this change but it's a catch 22 since it needs the cop to be updated which needs AST to be updated... 😅

dvandersluis:
I added that offense to `.rubocop_todo.yml` for now. I'll clean it up after everything is fixed.

(I didn't want to regenerate the whole file because it changes significantly and adds offenses to other files to remove `# rubocop:disable` directives -- we should maybe do this later though?)

marcandre:
Thanks for fixing this bug.

marcandre:
Released 1.33.1

dvandersluis:
Thanks @marcandre! ❤️

PR #326, opened 2024-11-04
Introduce `lhs` and `rhs` aliases to assignment nodes for consistency.

dvandersluis:
The methods are already available for `masgn`, so this will allow all assignment node types to have the same two methods so that they can be dealt with in unison. Once released I've found some places in rubocop where more destructuring can be simplified.

Follows https://github.com/rubocop/rubocop/pull/13413#discussion_r1827267164

I didn't add tests for these aliases but am happy to if desired!

### DIFF
--- a/lib/rubocop/ast/node/asgn_node.rb
+++ b/lib/rubocop/ast/node/asgn_node.rb
@@ -12,6 +12,7 @@ module RuboCop
       def name
         node_parts[0]
       end
+      alias lhs name
 
       # The expression being assigned to the variable.
       #
@@ -19,6 +20,7 @@ module RuboCop
       def expression
         node_parts[1]
       end
+      alias rhs expression
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/rubocop/ast/node/asgn_node.rb
```diff
@@ -12,13 +12,15 @@ class AsgnNode < Node
...
       end
+      alias lhs name
 
       # The expression being assigned to the variable.
       #
       # @return [Node] the expression being assigned.
       def expression
         node_parts[1]
       end
+      alias rhs name
```
bquorning, line 23:
```suggestion
      alias rhs expression
```

dvandersluis: Oops thanks @bquorning! Updated.

marcandre:
Looks good, thank you.

Released as 1.34.0.

Minor thing: If you think about it for upcoming PRs, please make a Changelog entry...

dvandersluis:
Sorry @marcandre! I'll add a changelog for this right now.

PR #327, opened 2024-11-04
Fix docs for `masgn`

dvandersluis:
There was a stray table cell in the `masgn` row which pushed everything below it off by a cell.


marcandre:
Good catch, thanks!

PR #330, opened 2024-11-04
Extend `Node#const_name` to `casgn` nodes

dvandersluis:
Now that `CasgnNode` was added in #204, we can extend `const_name` to `casgn` nodes as well. This allows code such as https://github.com/rubocop/rubocop/blob/736e8d699b5f632e7a05fd65788e28f23b2e3a67/lib/rubocop/cop/style/swap_values.rb#L84-L89 to be refactored out.


marcandre:
Totally. Thanks ❤️

PR #332, opened 2024-11-07
[Fix #331] Fix crash in `OpAsgnNode#name` when the lhs is a `send` or `csend` node

### LINKED ISSUE #331: OpAsgnNode doesn't work with explict receiver on the assignment node
Adding this example to spec/rubocop/ast/op_asgn_node_spec.rb should work fine, but the last expectation fails. 

```rb
describe 'assignment node with explicit receiver' do
  subject { op_asgn_node }

  let(:source) { 'self.var += value '}

  it 'behaves as expected' do
    expect(subject.class).to eq(RuboCop::AST::OpAsgnNode)
    expect(subject.assignment_node.source).to eq('self.var')
    expect(subject.operator).to eq(:+)
    expect(subject.expression.source).to eq('value')
    expect(subject.name).not_to raise_error
  end
end
```

The assignment_node is an instance of `RuboCop::AST::SendNode`, but the code assumes that it’s an instance of `RuboCop::AST::AsgnNode`, so an exception is raised when we call `assigment_node.name` in https://github.com/rubocop/rubocop-ast/blob/62d4112eae32f00c1e74c12480f2dbee00a81ff3/lib/rubocop/ast/node/op_asgn_node.rb#L19

cc @dvandersluis

dvandersluis:
`OpAsgnNode#assignment_node` could be a `send`/`csend` node, but this would cause `#name` to crash as call nodes do not respond to `name`.

Fixes #331.

### DIFF
--- a/changelog/fix_fix_331_fix_crash_in_opasgnnodename_when.md
+++ b/changelog/fix_fix_331_fix_crash_in_opasgnnodename_when.md
@@ -1,0 +1,1 @@
+* [#331](https://github.com/rubocop/rubocop-ast/issue/331): [Fix #331] Fix crash in `OpAsgnNode#name` when the lhs is a `send` or `csend` node. ([@dvandersluis][])
--- a/lib/rubocop/ast/node/op_asgn_node.rb
+++ b/lib/rubocop/ast/node/op_asgn_node.rb
@@ -16,7 +16,7 @@ module RuboCop
       #
       # @return [Symbol] the name of the variable being assigned
       def name
-        assignment_node.name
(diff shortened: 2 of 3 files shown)


dvandersluis:
@bquorning note that `name` isn't necessarily the source of the LHS (so your workaround in rubocop/rubocop#13426 might still be necessary). Similarly to #203, `name` will return the variable name, not the entire receiver.

bquorning (approved):

marcandre:
Thanks to both of you ❤️

Released as 1.34.1

PR #333, opened 2024-11-07
Add `BeginKeywordNode` for `kwbegin` nodes

dvandersluis:
Allows for `kwbegin` nodes to not need destructuring, as in rubocop/rubocop#13426.


dvandersluis:
@marcandre @bquorning this isn't quite as straightforward as I originally thought. 😅 

When the body of a `begin` is multiple lines, they are expressed as multiple items within the node:

```
$ ruby-parse -e '
begin
  foo
  bar
end
'
(kwbegin
  (send nil :foo)
  (send nil :bar))
```

Rubocop seems to act as if the `kwbegin` body is just the first one, but should it return an array of nodes instead?

I also added test cases for `begin...rescue...ensure...end` but I don't have it fully working yet.

dvandersluis:
I've also gone back and forth on if this should be `BeginNode`, `BeginKeywordNode`, or `KwbeginNode`. I've landed on `BeginKeywordNode` (because `begin` nodes also exist and could theoretically have a Node class one day) but let me know your thoughts.

marcandre:
Without much thought, I'd `body` should be an array, and name should be `KeywordBeginNode`.

dvandersluis:
> name should be KeywordBeginNode.

Ah yes of course, a fourth option I didn't consider 😂 I'll change it up this afternoon.

lib/rubocop/ast/node/keyword_begin_node.rb
```diff
@@ -0,0 +1,44 @@
...
+      # multiple nodes.
+      #
+      # @return [Node, nil] The body of the `kwbegin`.
+      def body
+        return unless node_parts.any?
+
+        if rescue_node
+          rescue_node.body
+        elsif ensure_node
+          ensure_node.node_parts[0]
```
dvandersluis, line 19: I'd like to do this more cleanly, but it's dependent on #334.

marcandre (approved):
Looks good to me 👍

lib/rubocop/ast/node/keyword_begin_node.rb
marcandre: Understood. I'd like to wait for the upcoming major release, though, so I'd say let's merge this now and tweak as part of #334

lib/rubocop/ast/node/ensure_node.rb
```diff
@@ -13,6 +13,13 @@ def body
         node_parts[1]
       end
 
+      # Returns the `rescue` node of the `ensure`, if present.
+      #
+      # @return [Node, nil] The `rescue` node.
+      def rescue_node
+        node_parts[0] if node_parts[0].rescue_type?
+      end
```
dvandersluis, line 21: @marcandre do you want me to do this as a separate PR, or create a separate changelog entry for it pointing to this PR?

marcandre:
Great, thanks. Released as 1.35

dvandersluis:
Thanks @marcandre!

PR #336, opened 2024-11-11
[#Fix 334] Make `body` consistent for `RescueNode`, `EnsureNode` and `KeywordBeginNode`

dvandersluis:
Renames `EnsureNode#body` to `#branch`, and adds a new `#body` method in line with `RescueNode` and `KeywordBeginNode` to return the body of code being executed before exception handling. For all three node types, `body` will now return a single node (which may be a `begin`/`kwbegin` node containing multiple nodes).

```ruby
begin
  foo       <--- body
  bar       <--- body
rescue 
  baz
ensure
  quux
end
```


dvandersluis:
I'm not sure when the plan for the next major version is, but getting this kicked off now so it's ready for whenever that'll be.

lib/rubocop/ast/node/ensure_node.rb
```diff
@@ -6,18 +6,27 @@ module AST
...
       def body
+        return rescue_node.body if rescue_node
+
+        node_parts[0]
+      end
+
+      # Returns an the ensure branch in the exception handling statement.
+      #
+      # @return [Node, nil] the body of the ensure branch.
+      def branch
```
dvandersluis, line 21: I used `branch` for the method to get the ensure branch body, similiar to how we have `branches` for `RescueNode`.

dvandersluis:
I've gone through every rubocop-* repo in the @rubocop org. As far as I can see, only rubocop/rubocop and rubocop/rubocop-minitest are affected by this change. For tracking purposes, here are the PRs to update `EnsureNode.body` in those repos:

* rubocop/rubocop#13456
* rubocop/rubocop-minitest#324

PR #337, opened 2024-11-12
Deprecate `EnsureNode#body` in favour of `EnsureNode#branch`

dvandersluis:
Given the upcoming method change in #336, but also the fact that the tests of that PR cannot pass as-is since rubocop is still relying on the old definition, I thought it would make sense to deprecate `EnsureNode#body` and add `#branch` as an alias. Once released, I can update the uses of `EnsureNode#body` in rubocop (and extensions) without breaking anything, which will allow the "Main Gem Specs" CI to pass.

### DIFF
--- a/lib/rubocop/ast/node/ensure_node.rb
+++ b/lib/rubocop/ast/node/ensure_node.rb
@@ -6,10 +6,31 @@ module RuboCop
     # node when the builder constructs the AST, making its methods available
     # to all `ensure` nodes within RuboCop.
     class EnsureNode < Node
+      DEPRECATION_WARNING_LOCATION_CACHE = [] # rubocop:disable Style/MutableConstant
+      private_constant :DEPRECATION_WARNING_LOCATION_CACHE
+
       # Returns the body of the `ensure` clause.
       #
       # @return [Node, nil] The body of the `ensure`.
+      # @deprecated Use `EnsureNode#branch`
       def body
+        first_caller = caller(1..1).first
+
+        unless DEPRECATION_WARNING_LOCATION_CACHE.include?(first_caller)
+          warn '`EnsureNode#body` is deprecated and will be changed in the next major version of ' \
+               'rubocop-ast. Use `EnsureNode#branch` instead to get the body of the `ensure` branch.'
+          warn "Called from:\n#{caller.join("\n")}\n\n"
+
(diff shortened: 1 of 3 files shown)


lib/rubocop/ast/node/ensure_node.rb
```diff
@@ -6,10 +6,31 @@ module AST
...
+        unless DEPRECATION_WARNING_LOCATION_CACHE.include?(first_caller)
+          warn '`EnsureNode#body` is deprecated and will be changed in the next major version of ' \
+               'rubocop-ast. Use `EnsureNode#branch` instead to get the body of the `ensure` branch.'
+          warn "Called from:\n#{caller.join("\n")}\n\n"
+
+          DEPRECATION_WARNING_LOCATION_CACHE << first_caller
+        end
+
+        branch
+      end
```
dvandersluis, line 28: I'm not sure if this is overkill?

I thought it'd be useful to output a warning given that the method will no longer work the same post #336 release. At the same time, I am going to update all the @rubocop org repos to use the new method name, so the risk is only to 3rd party extensions that might use `EnsureNode#body`.

I also wanted it to not spam deprecation warnings, hence the cache to make sure each warning is just output once.

dvandersluis: @Earlopain it looks like `StrictWarnings` is causing the rubocop master specs to fail because of the warning 🤦. Any ideas?

Earlopain: Well, this is awkward. I didn't anticipate this interaction.

Currently there is nothing to disable this but it is easy enough to control through environment variable. It already does nothing if `CI` is not set but I don't think that would be appropriate here. I can make a PR tomorrow.

Rails sets `STRICT_WARNINGS=1` in their CI instead of relying on something generic, once again I should just have followed what they do.

marcandre: > I'm not sure if this is overkill?

I don't think so, I like it.

marcandre:
Great. Could you add a Changelog entry please 🙏

dvandersluis:
Yes! Updated now.

Earlopain:
One consideration, I don't believe this should show the deprecation warning rightaway. There is currently no RuboCop version that will not show this warning

Earlopain:
https://github.com/rubocop/rubocop-ast/pull/339

PR #339, opened 2024-11-13
Don't show a deprecation warning for `EnsureNode#branch` just yet

Earlopain:
There currently is no RuboCop version that will not show this warning.
Somehow this did not cross my mind while looking at the other PR...

This can be reverted after RuboCop and its extensions are updated and released  to use the non-deprecated method.

### DIFF
--- a/changelog/fix_dont_emit_deprecation_just_yet.md
+++ b/changelog/fix_dont_emit_deprecation_just_yet.md
@@ -1,0 +1,1 @@
+* [#339](https://github.com/rubocop/rubocop-ast/pull/339): Do not emit a deprecation warning for `EnsureNode#body` to give RuboCop a chance to update its usage. `EnsureNode#body` will still be changed in the next major version of `rubocop-ast`. ([@earlopain][])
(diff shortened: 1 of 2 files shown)


dvandersluis:
Shoot, you're right. I was thinking that rubocop wouldn't be affected until rubocop-ast was updated manually, but we're not actually pinning a version of rubocop-ast, so anyone who installs rubocop fresh or updates it will get a noisy version of rubocop-ast. 🤦 

@marcandre sorry I missed this too. Perhaps 1.36 should be yanked and this should be released as 1.36.1?

marcandre:
Right, clearly I lack experience in deprecating.
First soft deprecate, then update dependencies, then deprecate, then change...

dvandersluis:
Me too @marcandre. Thanks for your quick attention to this, and thanks @Earlopain for noticing!

Earlopain:
Doesn't happen to often. I would suggest giving 1 or 2 months after release with fixes to rubocop and similar to give people the chance to upgrade.

They might restrict rubocop in their gemfile but I don't think the same is common for `rubocop-ast`. Especially people on `standard` won't be able to do much about it as I've recently learned for unrelated reasons.

dvandersluis:
I think that's probably reasonable to add the warning back, although I haven't checked that the deprecated method hasn't had new uses added.

PR #340, opened 2024-11-27
Handle anonymous splats in `MlhsNode`

Earlopain:
Ref: https://github.com/rubocop/rubocop/issues/13505

I checked what the cop was previously expecting here, before https://github.com/rubocop/rubocop/pull/13422

I'm not sure if a `splat` makes much sense here, but nil seems even more unexpected.

cc @dvandersluis


dvandersluis (approved):
Makes sense to me.

marcandre:
Agreed, thanks!

Released as 1.36.2

PR #341, opened 2024-12-06
Make `RuboCop::AST::IfNode` support `then?`

koic:
This PR makes `RuboCop::AST::IfNode` support `then?`, similar to `RuboCop::AST::InPatternNode` and `RuboCop::AST::WhenNode`.

### DIFF
--- a/lib/rubocop/ast/node/if_node.rb
+++ b/lib/rubocop/ast/node/if_node.rb
@@ -25,6 +25,13 @@ module RuboCop
         keyword == 'unless'
       end
 
+      # Checks whether the `if` node has an `then` clause.
+      #
+      # @return [Boolean] whether the node has an `then` clause
+      def then?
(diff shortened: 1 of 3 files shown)


lib/rubocop/ast/node/if_node.rb
```diff
@@ -25,6 +25,13 @@ def unless?
         keyword == 'unless'
       end
 
+      # Checks whether the `if` node has an `then` clause.
+      #
+      # @return [Boolean] whether the node has an `then` clause
+      def then?
+        loc.begin&.is?('then')
```
marcandre, line 32: Could you please rebase and use the new `loc_is?` here? 🙏

koic: Sure! I've updated it!

marcandre:
Thank you!

PR #343, opened 2024-12-12
Add `StrNode#single_quoted?`, `StrNode#double_quoted?` and `StrNode#percent_literal?` to simplify checking for string delimiters

dvandersluis:
This would allow rubocop code that checks for different string delimiter types to do so in a simpler way.

I took inspiration from `ArrayNode#percent_literal?` for the interface for `StrNode#percent_literal?`.

### DIFF
--- a/lib/rubocop/ast/node/str_node.rb
+++ b/lib/rubocop/ast/node/str_node.rb
@@ -8,12 +8,48 @@ module RuboCop
     class StrNode < Node
       include BasicLiteralNode
 
+      PERCENT_LITERAL_TYPES = {
+        :% => /\A%(?=[^a-zA-Z])/,
+        :q => /\A%q/,
+        :Q => /\A%Q/
+      }.freeze
+
+      def single_quoted?
+        loc_is?(:begin, "'")
+      end
+
+      def double_quoted?
+        loc_is?(:begin, '"')
+      end
+
       def character_literal?
         loc_is?(:begin, '?')
       end
 
       def heredoc?
         loc.is_a?(Parser::Source::Map::Heredoc)
+      end
+
+      # Checks whether the string literal is delimited by percent brackets.
+      #
+      # @overload percent_literal?
+      #   Check for any string percent literal.
+      #
+      # @overload percent_literal?(type)
+      #   Check for a string percent literal of type `type`.
+      #
+      # @param type [Symbol] an optional percent literal type
+      #
+      # @return [Boolean] whether the string is enclosed in percent brackets
+      def percent_literal?(type = nil)
+        opening_delimiter = loc.begin if loc.respond_to?(:begin)
+        return false unless opening_delimiter
+
+        if type
+          opening_delimiter.source.match?(PERCENT_LITERAL_TYPES.fetch(type))
+        else
+          opening_delimiter.source.start_with?('%')
+        end
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/rubocop/ast/node/str_node.rb
```diff
@@ -8,13 +8,56 @@ module AST
     class StrNode < Node
       include BasicLiteralNode
 
+      PERCENT_LITERAL_TYPES = {
+        :% => /^%(?![qQ])/,
+        :q => /^%q/,
+        :Q => /^%Q/
```
koic, line 14: It seems that matching the beginning of lines in a multiline string is not expected.

```suggestion
        :% => /\A%(?![qQ])/,
        :q => /\A%q/,
        :Q => /\A%Q/
```

lib/rubocop/ast/node/str_node.rb
```diff
@@ -8,13 +8,56 @@ module AST
     class StrNode < Node
       include BasicLiteralNode
 
+      PERCENT_LITERAL_TYPES = {
+        :% => /^%(?![qQ])/,
```
Earlopain, line 12: `%%foo%`, probably to the horror of some, is a valid string. I'm not sure if any cop currently cares, just wanted to point it out. I'm not 100% on what would be a valid delimiter for these, maybe `[^a-zA-Z]` would be good enough?

dvandersluis: That'd still be a `%(...)` string though, wouldn't it? The extra %s replace brackets but the macro type is still a single %.

Earlopain: Yes, I think so. It's in a similar vein as `%w!a!` being a valid array (hopefully no one actually writes such code).

dvandersluis:
Thanks @koic and @Earlopain, the requested changes have been made.

lib/rubocop/ast/node/str_node.rb
```diff
@@ -8,13 +8,56 @@ module AST
...
+      #   Check for a string percent literal of type `type`.
+      #
+      # @param type [Symbol] an optional percent literal type
+      #
+      # @return [Boolean] whether the string is enclosed in percent brackets
+      def percent_literal?(type = nil)
+        return false unless opening_delimiter
+
+        if type
+          opening_delimiter.source.match?(PERCENT_LITERAL_TYPES[type])
```
marcandre, line 48: Let's use `fetch` for a better error message in case `type` doesn't have an expected value

lib/rubocop/ast/node/str_node.rb
```diff
@@ -8,13 +8,56 @@ module AST
...
+        else
+          opening_delimiter.source.start_with?('%')
+        end
+      end
+
+      private
+
+      def opening_delimiter
+        return unless loc.respond_to?(:begin)
```
marcandre, line 57: Is it possible that a `str` node's location does not respond to begin?

marcandre: (We could use `loc_is?` that I propose in #345)

dvandersluis: > Is it possible that a `str` node's location does not respond to begin?

Yes, there are `str` nodes without delimiters inside other nodes (such as `dstr` or `regexp`, not an exhaustive list) that have no delimiters and thus no `loc.begin`. I added a test to cover this.

dvandersluis: > (We could use `loc_is?` that I propose in #345)

I like it!

lib/rubocop/ast/node/str_node.rb
dvandersluis: Updated.

lib/rubocop/ast/node/str_node.rb
dvandersluis: Updated to use `loc_is?` now.

PR #344, opened 2024-12-12
`reminds` probably was meant to be `resembles`

Kallin:
I believe there's a typo in the `node_pattern` docs. This change fixes the sentence.

### DIFF
--- a/docs/modules/ROOT/pages/node_pattern.adoc
+++ b/docs/modules/ROOT/pages/node_pattern.adoc
@@ -3,7 +3,7 @@
 Node pattern is a DSL to help find specific nodes in the Abstract Syntax Tree
 using a simple string.
 
-It reminds the simplicity of regular expressions but used to find specific
+It resembles the simplicity of regular expressions but used to find specific
 nodes of Ruby code.
 
 == History
--- a/docs/modules/ROOT/pages/node_pattern.adoc
+++ b/docs/modules/ROOT/pages/node_pattern.adoc
@@ -3,7 +3,7 @@
 Node pattern is a DSL to help find specific nodes in the Abstract Syntax Tree
 using a simple string.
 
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/node_pattern.adoc
```diff
@@ -3,7 +3,7 @@
 Node pattern is a DSL to help find specific nodes in the Abstract Syntax Tree
 using a simple string.
 
-It reminds the simplicity of regular expressions but used to find specific
+It resembles the simplicity of regular expressions but used to find specific
```
Earlopain, line 6: I guess it was supposed to be this?

```suggestion
It reminds of the simplicity of regular expressions but used to find specific
```

Kallin: `It reminds of the` is not clear English. 
Adding the 2nd `is` does make sense though.

Kallin, line 6:
```suggestion
It resembles the simplicity of regular expressions but is used to find specific
```

Kallin: updated the PR w/ the change

Kallin: You might say:

`It reminds one of the simplicity..` or `It may remind you of the simplicity`... but that sounds too informal.

Kallin:
even better:

```
It evokes the pattern-matching capabilities of regular expressions, but is designed to target specific nodes in Ruby code.
```

marcandre:
> even better:
> 
> ```
> It evokes the pattern-matching capabilities of regular expressions, but is designed to target specific nodes in Ruby code.
> ```

That's nice, I went for that and put you as author @Kallin , thanks

PR #347, opened 2024-12-13
Refactor `ArrayNode#percent_literal?`

dvandersluis:
I used the feedback on #343 to make similar changes to `ArrayNode#percent_literal?` as were suggested for `StrNode#percent_literal?`.

### DIFF
--- a/lib/rubocop/ast/node/str_node.rb
+++ b/lib/rubocop/ast/node/str_node.rb
@@ -13,6 +13,7 @@ module RuboCop
         :q => /\A%q/,
         :Q => /\A%Q/
       }.freeze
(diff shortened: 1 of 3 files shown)


lib/rubocop/ast/node/str_node.rb
```diff
@@ -13,6 +13,7 @@ class StrNode < Node
         :q => /\A%q/,
         :Q => /\A%Q/
       }.freeze
+      private_constant :PERCENT_LITERAL_TYPES
```
dvandersluis, line 16: When looking at `ArrayNode` I noticed I forgot to mark this constant as private for `StrNode` so I threw that in here too.

PR #349, opened 2024-12-28
Better handle `LoadError`s from prism

Earlopain:
There are two different cases that may happen:

Prism was compiled in the past, the user updated the ruby version, but it still points to the same library
> linked to incompatible /home/user/.rbenv/versions/3.4.1/lib/libruby.so.3.4 - /home/user/Documents/ruby-prism/lib/prism/prism.so

In that case, bubble up the original error.

On the other hand, since prism is no dependency, users may run older verions that are missing translator classes. In that case, show a message what they should do.
prism is a default gem since 3.3, so you can even use it without adding it to your gemfile.

### DIFF
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -318,7 +318,7 @@ module RuboCop
         require "prism/translation/parser#{version.to_s.delete('.')}"
       rescue LoadError
         warn <<~MESSAGE
-          Error: Unable to load Prism Parser for Ruby #{version}.
+          Error: Unable to load Prism parser for Ruby #{version}.
           * If you're using Bundler and don't yet have `gem 'prism'` as a dependency, add it now.
           * If you're using Bundler and already have `gem 'prism'` as a dependency, update it to the most recent version.
           * If you don't use Bundler, run `gem update prism`.
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -285,25 +285,45 @@ module RuboCop
             raise ArgumentError, "RuboCop found unknown Ruby version: #{ruby_version.inspect}"
           end
         when :parser_prism
-          begin
-            require 'prism'
-          rescue LoadError
-            warn "Error: Unable to load Prism. Add `gem 'prism'` to your Gemfile."
-            exit!
-          end
+          require_prism
 
           case ruby_version
           when 3.3
-            require 'prism/translation/parser33'
+            require_prism_translation_parser(ruby_version)
             Prism::Translation::Parser33
           when 3.4
-            require 'prism/translation/parser34'
+            require_prism_translation_parser(ruby_version)
             Prism::Translation::Parser34
           else
             raise ArgumentError, 'RuboCop supports target Ruby versions 3.3 and above with Prism. ' \
                                  "Specified target Ruby version: #{ruby_version.inspect}"
           end
         end
+      end
+
+      # Prism is a native extension, a `LoadError` will be raised if linked to an incompatible
+      # Ruby version. Only raise if it really was caused by Prism not being present.
+      def require_prism
+        require 'prism'
+      rescue LoadError => e
+        raise unless e.path == 'prism'
+
+        warn "Error: Unable to load Prism. Add `gem 'prism'` to your Gemfile."
+        exit!
+      end
+
+      # While Prism is not yet a dependency, users may run with outdated versions that
+      # don't have all the parsers.
+      def require_prism_translation_parser(version)
+        require "prism/translation/parser#{version.to_s.delete('.')}"
+      rescue LoadError
+        warn <<~MESSAGE
+          Error: Unable to load Prism Parser for Ruby #{version}.
+          * If you're using Bundler and don't yet have `gem 'prism'` as a dependency, add it now.
+          * If you're using Bundler and already have `gem 'prism'` as a dependency, update it to the most recent version.
+          * If you don't use Bundler, run `gem update prism`.
+        MESSAGE
+        exit!
       end
 
       def create_parser(ruby_version, parser_engine)


lib/rubocop/ast/processed_source.rb
```diff
@@ -306,6 +301,31 @@ def parser_class(ruby_version, parser_engine) # rubocop:disable Metrics/AbcSize,
...
+      rescue LoadError => e
+        raise unless e.path == 'prism'
+
+        warn "Error: Unable to load Prism. Add `gem 'prism'` to your Gemfile."
+        exit!
+      end
+
+      # While prism is not yet a dependency, users may run with outdated versions that
+      # don't have all the parsers.
+      def require_prism_parser(version)
```
koic, line 317:
```suggestion
      def require_prism_translation_parser(version)
```

lib/rubocop/ast/processed_source.rb
```diff
@@ -306,6 +301,31 @@ def parser_class(ruby_version, parser_engine) # rubocop:disable Metrics/AbcSize,
         end
       end
 
+      # Prism is a native extension, a LoadError will be raised if linked to an incompatible
+      # Ruby version. Only raise if this NameError really was caused by prism not being present.
```
koic, line 305:
```suggestion
      # Prism is a native extension, a `LoadError` will be raised if linked to an incompatible
      # Ruby version. Only raise if this `NameError` really was caused by Prism not being present.
```

lib/rubocop/ast/processed_source.rb
```diff
@@ -306,6 +301,31 @@ def parser_class(ruby_version, parser_engine) # rubocop:disable Metrics/AbcSize,
...
+      def require_prism
+        require 'prism'
+      rescue LoadError => e
+        raise unless e.path == 'prism'
+
+        warn "Error: Unable to load Prism. Add `gem 'prism'` to your Gemfile."
+        exit!
+      end
+
+      # While prism is not yet a dependency, users may run with outdated versions that
```
koic, line 315:
```suggestion
      # While Prism is not yet a dependency, users may run with outdated versions that
```

lib/rubocop/ast/processed_source.rb
```diff
@@ -306,6 +301,31 @@ def parser_class(ruby_version, parser_engine) # rubocop:disable Metrics/AbcSize,
...
+      # don't have all the parsers.
+      def require_prism_parser(version)
+        require "prism/translation/parser#{version.to_s.delete('.')}"
+      rescue LoadError
+        warn <<~WARN
```
koic, line 320:
```suggestion
        warn <<~MESSAGE
```

or

```diff
-        warn <<~WARN
+        warn <<~WARNING
```

changelog/fix_better_handle_prism_loaderror.md
```diff
@@ -0,0 +1 @@
+* [#349](https://github.com/rubocop/rubocop-ast/pull/349): Show better error messages when `prism` is too old or linked to the wrong Ruby version. ([@earlopain][])
```
koic, line 1: Since the behavior of the feature for users doesn't change, the changelog entry is likely unnecessary. Can you remove it?

lib/rubocop/ast/processed_source.rb
```diff
@@ -306,6 +301,31 @@ def parser_class(ruby_version, parser_engine) # rubocop:disable Metrics/AbcSize,
...
+          Error: Unable to load Prism Parser for Ruby #{version}.
+          * If you're using bundler and don't yet have prism as a dependency, add it now.
+          * If you're using bundler and already have prism as a dependency, update it to the most recent version.
```
koic, line 323:
```suggestion
          * If you're using Bundler and don't yet have `gem prism` as a dependency, add it now.
          * If you're using Bundler and already have `gem 'prism'` as a dependency, update it to the most recent version.
```

or

```diff
-          * If you're using bundler and don't yet have prism as a dependency, add it now.
-          * If you're using bundler and already have prism as a dependency, update it to the most recent version.
+          * If you're using Bundler and don't yet have Prism as a dependency, add it now.
+          * If you're using Bundler and already have Prism as a dependency, update it to the most recent version.
```

lib/rubocop/ast/processed_source.rb
```diff
@@ -306,6 +301,31 @@ def parser_class(ruby_version, parser_engine) # rubocop:disable Metrics/AbcSize,
...
+          * If you don't use bundler, run `gem update prism`.
```
koic, line 324:
```suggestion
          * If you don't use Bundler, run `gem update prism`.
```

lib/rubocop/ast/processed_source.rb
```diff
@@ -285,19 +285,14 @@ def parser_class(ruby_version, parser_engine) # rubocop:disable Metrics/AbcSize,
...
-          rescue LoadError
-            warn "Error: Unable to load Prism. Add `gem 'prism'` to your Gemfile."
-            exit!
-          end
+          require_prism
 
           case ruby_version
           when 3.3
-            require 'prism/translation/parser33'
+            require_prism_parser(3.3)
```
koic, line 292:
```suggestion
            require_prism_parser(ruby_version)
```

lib/rubocop/ast/processed_source.rb
```diff
@@ -285,19 +285,14 @@ def parser_class(ruby_version, parser_engine) # rubocop:disable Metrics/AbcSize,
...
             Prism::Translation::Parser33
           when 3.4
-            require 'prism/translation/parser34'
+            require_prism_parser(3.4)
```
koic, line 295:
```suggestion
            require_prism_parser(ruby_version)
```

Earlopain:
Thanks, I think I've applied all your suggestions

lib/rubocop/ast/processed_source.rb
```diff
@@ -306,6 +301,31 @@ def parser_class(ruby_version, parser_engine) # rubocop:disable Metrics/AbcSize,
...
+        exit!
+      end
+
+      # While Prism is not yet a dependency, users may run with outdated versions that
+      # don't have all the parsers.
+      def require_prism_translation_parser(version)
+        require "prism/translation/parser#{version.to_s.delete('.')}"
+      rescue LoadError
+        warn <<~MESSAGE
+          Error: Unable to load Prism Parser for Ruby #{version}.
```
koic, line 321: That's just my two cents.

```suggestion
          Error: Unable to load Prism parser for Ruby #{version}.
```

marcandre:
Thanks to both of you ❤️

PR #353, opened 2024-12-28
Clean up CI a bit

Earlopain:
Let's see if I got this right on the first try :upside_down_face: (I didn't)

A few months ago I and @\bquorning made some tweaks to rubocops CI. This imports some of that work, and does a bit specific to this repo.

I made a handful of individual commits explaining why I did something for each. Overall the workflow files lost a bit of weight and are easier to follow.


marcandre:
Awesome, thanks!

PR #354, opened 2025-01-21
Prism custom builder

Earlopain:
Context: https://github.com/ruby/prism/pull/3443

I'd like for prism to emit ast for new ruby features, like `it`. It's incredibly easy for prism to emit `itblock`, it just needs a custom builder, with a few lines changed to do so. But since `rubocop-ast` already uses its own custom builder, they need to play nicely together for it to work.

The PR for prism just adds the builder class, without any changes. I'd like to explore that after this.

### DIFF
--- a/changelog/change_prism_builder.md
+++ b/changelog/change_prism_builder.md
@@ -1,0 +1,1 @@
+* [#354](https://github.com/rubocop/rubocop-ast/pull/354): Use `Prism::Translation::Parser::Builder` when parsing with prism. ([@earlopain][])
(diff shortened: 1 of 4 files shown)


Earlopain:
This is ready now, prism 1.4.0 got released which contains the necessary changes. One main rubocop spec currently fails for unrelated reasons (https://github.com/rubocop/rubocop/issues/14000).

For a bit of context, in https://github.com/rubocop/rubocop/issues/13617#issuecomment-2734551361 I wrote what I believe should be the next steps.

marcandre:
Cool, thanks. Released as v1.40

PR #355, opened 2025-01-21
Test main specs against prism as well

Earlopain:
Makes sense, I think. Especially relevant if the prism translator gains functionality that the normal parser one doesn't have


marcandre:
Cool, thanks

PR #356, opened 2025-01-21
Add `Node#any_block_type?` to determine if a node is either a block or numblock

Earlopain:
There are many instances where you just want to know if it's a block and don't care about what type it is: https://github.com/search?q=org%3Arubocop+%2Fblock.*numblock%2F&type=code

Should `itblock` be added, most of these will need to add yet another block type. With this, they can instead use this.

I had a bit of trouble choosing a good name. Maybe you have a better one to use?

### DIFF
--- a/changelog/new_group_block.md
+++ b/changelog/new_group_block.md
@@ -1,0 +1,1 @@
+* [#356](https://github.com/rubocop/rubocop-ast/pull/356): Added `:any_block` as an alias for `:block` and `:numblock`, use it with `Node#any_block_type?`. Also available in node patterns: `{block numblock}` can become `any_block`. ([@earlopain][])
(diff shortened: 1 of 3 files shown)


koic:
This was originally envisioned as `gen_block`, but I think `any_block` is a sufficiently understandable naming choice. https://github.com/rubocop/rubocop-ast/pull/205#issuecomment-904313695

Especially, `node.any_block_type?` is likely easy to read and understand 👍

marcandre:
I was also going to say `any_block` 👍

changelog/new_group_block.md
```diff
@@ -0,0 +1 @@
+* [#356](https://github.com/rubocop/rubocop-ast/pull/356): Add `Node#any_block_type?` to determine if a node is either a block or numblock. ([@earlopain][])
```
marcandre, line 1: I would start by saying that an alias `:any_block` was added for `block` and `numblock` and then mention the method

PR #357, opened 2025-01-21
Support node groups in `Node#each_descendant` and similar traversal methods

Earlopain:
Sometimes it would be nice to do `each_child_node(:call)` instead of `each_child_node(:send, :csend)`. This nicely mirrors that you can already use these aliases when writing node patterns

### DIFF
--- a/lib/rubocop/ast/node.rb
+++ b/lib/rubocop/ast/node.rb
@@ -76,9 +76,6 @@ module RuboCop
       OPERATOR_KEYWORDS = %i[and or].to_set.freeze
       # @api private
       SPECIAL_KEYWORDS = %w[__FILE__ __LINE__ __ENCODING__].to_set.freeze
-      # @api private
-      ARGUMENT_TYPES = %i[arg optarg restarg kwarg kwoptarg kwrestarg
-                          blockarg forward_arg shadowarg].to_set.freeze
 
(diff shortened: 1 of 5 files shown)


lib/rubocop/ast/node.rb
```diff
@@ -98,7 +95,7 @@ class Node < Parser::AST::Node # rubocop:disable Metrics/ClassLength
         kwrestarg: :argument,
         blockarg: :argument,
         forward_arg: :argument,
-        shardowarg: :argument,
+        shadowarg: :argument,
```
Earlopain, line 98: Typo since this was added, tests caught it when I started using `each_descendant(:argument)`

lib/rubocop/ast/node.rb
```diff
@@ -528,6 +525,14 @@ def range_type?
         GROUP_FOR_TYPE[type] == :range
       end
 
+      def type_or_group_type?(type)
+        type == self.type || GROUP_FOR_TYPE[self.type] == type
+      end
```
koic, line 530: I'm uncertain whether the behavior indicated by the name `type_or_group_type?` is clear. Since it currently seems to be called only from `yield_for_types`, how about not publishing `type_or_group_type?` and making it a private method instead?

Earlopain: I took another look and seems like I just reimplemented the existing `type?` method. So, I removed both of these new methods instead.

marcandre:
Totally! ❤️
Thanks

PR #358, opened 2025-01-22
[Fix #322] Fix node captures inside of `?`, `+`, and `*` repetition

### LINKED ISSUE #322: `?` repetition causes matched nodes to be returned as an array
https://nodepattern.herokuapp.com/?p=(%0A%20%20def%20%24_method_name%0A%20%20(args)%0A%20%20(or-asgn%20%24(ivasgn%20_%20...)%20...)%3F%0A)&ruby=def%20foo%0A%20%20%40bar%20%7C%7C%3D%20baz%0Aend

grafik

The pattern is reduced from what I'm trying to write. I'd like to use OR `{}` but because each branch will contain different amount of captures, I write multiple `?` matchers after each other. `*` and `+` can return 0-n and 1-n matches so it totally makes sense to return an array for these. `+` on the other hand is ony 0-1 so I'd expect this to be either be the node or `nil`.

Can't change this now without a major version bump of course and the solution for me is to just call `flatten` somewhere in the code but I wanted to see what others think. Does my "complaint" make sense?

Earlopain:
Fix https://github.com/rubocop/rubocop-ast/issues/323

The calculation for how many captures a pattern has was not correct in this case. A pattern like `(send _ _ (int {$_ | $_ | 
$_})?)` would end up with 3 captures, even though the union would in total only contribute one

### DIFF
--- a/changelog/fix_captures_in_repetition.md
+++ b/changelog/fix_captures_in_repetition.md
@@ -1,0 +1,1 @@
+* [#323](https://github.com/rubocop/rubocop-ast/issues/323): Fix node captures inside of `?`, `+`, and `*` repetition. ([@earlopain][])
--- a/lib/rubocop/ast/node_pattern/node.rb
+++ b/lib/rubocop/ast/node_pattern/node.rb
@@ -48,7 +48,7 @@ module RuboCop
           children[0]
         end
 
-        # @return [Integer] nb of captures of that node and its descendants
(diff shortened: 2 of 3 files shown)


marcandre:
Great catch! 🚀
Could you refine the function for the `Union` class instead though? 🙏

Earlopain:
Much better, thanks for the suggestion :+1:

marcandre:
Great, thanks!

PR #361, opened 2025-02-25
[Fix #360] Fix an error when the `Array` core class contains a writer method

### LINKED ISSUE #360: faster is not better when it booms
`SimpleForwardable` is fast but not good.

```sh
drawohara@drawohara.dev:~ #=> cat faster-is-not-better-when-it-goes-boom.rb 
```

```ruby
module RuboCop
  # Similar to `Forwardable#def_delegators`, but simpler & faster
  module SimpleForwardable
    def def_delegators(accessor, *methods)
      methods.each do |method|
        class_eval(<<~RUBY, __FILE__, __LINE__ + 1)
          def #{method}(...)           # def example(...)
            #{accessor}.#{method}(...) #   foo.example(...)
          end                          # end
        RUBY
      end
    end
  end
end

class C
  def to_a
    []
  end

  extend RuboCop::SimpleForwardable

  def_delegators :to_a, :fields=
end
```

```sh
drawohara@drawohara.dev:~ #=> ruby faster-is-not-better-when-it-goes-boom.rb 
faster-is-not-better-when-it-goes-boom.rb:6:in `class_eval': 
faster-is-not-better-when-it-goes-boom.rb:8: syntax error, unexpected ')' (SyntaxError)
  to_a.fields=(...) #   foo.example(...)
                  ^

        from faster-is-not-better-when-it-goes-boom.rb:6:in `block in def_delegators'
        from faster-is-not-better-when-it-goes-boom.rb:5:in `each'
        from faster-is-not-better-when-it-goes-boom.rb:5:in `def_delegators'
        from faster-is-not-better-when-it-goes-boom.rb:23:in `<class:C>'
        from faster-is-not-better-when-it-goes-boom.rb:16:in `<main>'
```

Earlopain:
Fix #360

This is a syntax error:
```
def foo=(...)
end
```

So it must use `public_send` instead in that case. It's caused by https://github.com/rubocop/rubocop-ast/blob/47e0eead4b84c3d28d7251a548e152f914219773/lib/rubocop/ast/node/mixin/collection_node.rb#L9-L10

By default the `Array` class doesn't contain such methods but when something like spring is in use, code can be evaluated before rubocop.

I thought about simply rejecting such methods but supporting it is easy enough

### DIFF
--- a/changelog/fix_array_writer_method.md
+++ b/changelog/fix_array_writer_method.md
@@ -1,0 +1,1 @@
+* [#360](https://github.com/rubocop/rubocop-ast/pull/360): Fix an error when the `Array` core class contains a writer method before `rubocop-ast` loaded. ([@earlopain][])
--- a/lib/rubocop/ast/utilities/simple_forwardable.rb
+++ b/lib/rubocop/ast/utilities/simple_forwardable.rb
@@ -5,11 +5,22 @@ module RuboCop
   module SimpleForwardable
     def def_delegators(accessor, *methods)
       methods.each do |method|
-        class_eval(<<~RUBY, __FILE__, __LINE__ + 1)
-          def #{method}(...)           # def example(...)
-            #{accessor}.#{method}(...) #   foo.example(...)
-          end                          # end
-        RUBY
+        if method.end_with?('=') && method.to_s != '[]='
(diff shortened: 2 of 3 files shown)


marcandre:
Thanks!

PR #363, opened 2025-02-27
[Fix #348] Automate the process of GitHub release creation

### LINKED ISSUE #348: GitHub releases are missing
Just notice that:

https://github.com/rubocop/rubocop-ast/releases

Latest release [rubocop-ast 0.1.0](https://github.com/rubocop/rubocop-ast/releases/tag/v0.1.0), which is misleading

Probably need to add subtask to CI release process also cut a github release, like it works in main rubocop repo

koic:
Follow-up to https://github.com/rubocop/rubocop/blob/master/.github/workflows/github_release.yml.

This action will be triggered when a new tag is pushed and will auto-fill the release notes using the relevant file. The `rubocop-ast` repository has essentially the same structure as the `rubocop` repository, so it should work as is.

Resolves #348.


.github/workflows/github_release.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+
+      - name: Create GitHub Release
+        uses: ncipollo/release-action@v1
+        with:
+          tag: ${{ github.ref_name }}
+          name: RuboCop AST ${{ github.ref_name }}
+          bodyFile: relnotes/${{ github.ref_name }}.md
```
Earlopain, line 24: This won't work, there's no directory like this here. You would need to import the code that generates this from rubocop as well

bbatsov: The simplest workaround is to dump the entire changelog there, but it won't be as nice.

koic: Yeah, I created a `relnotes` file and included past release files to make this workflow functional. Going forward, this workflow can work based on `relnotes`.

bbatsov: Great! It's be nice if we use the same approach across all official RuboCop projects.

PR #364, opened 2025-03-11
Modify `def_callback` code to be easier to understand

Earlopain:
I had a bit of trouble following how this code actually works.

I find the argument names to be unfortunatly named. When I see signature/arity, I think of methods. But it actually refers to the structure of child nodes here.

There also was the `arity_check` argument, which isn't necessary. Reusing the arity with a endless range makes it a bit simpler.

### DIFF
--- a/lib/rubocop/ast/traversal.rb
+++ b/lib/rubocop/ast/traversal.rb
@@ -26,16 +26,18 @@ module RuboCop
         SEND = 'send(TYPE_TO_METHOD[child.type], child)'
         assign_code = 'child = node.children[%<index>i]'
         code = "#{assign_code}\n#{SEND}"
+        # How a particular child node should be visited. For example, if a child node
+        # can be nil it should be guarded behind a nil check. Or, if a child node is a literal
+        # (like a symbol) then the literal itself should not be visited.
         TEMPLATE = {
           skip: '',
           always: code,
           nil?: "#{code} if child"
         }.freeze
 
-        def def_callback(type, *signature,
-                         arity: signature.size..signature.size,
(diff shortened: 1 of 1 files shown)


lib/rubocop/ast/traversal.rb
```diff
@@ -72,18 +81,18 @@ def arity_check(range)
       extend CallbackCompiler
       send_code = CallbackCompiler::SEND
 
-      ### arity == 0
+      ### children == 0
```
marcandre, line 84: Could we write `0 children` or `children count == 0` (and same with subsequent comments)

lib/rubocop/ast/traversal.rb
```diff
@@ -49,16 +51,23 @@ def on_#{type}(node)        # def on_send(node)
...
             .join("\n")
+
+          <<~BODY
+            #{children_check_code}
+            #{visit_children_code}
+          BODY
         end
 
-        def arity_check(range)
+        def children_check(range)
```
marcandre, line 70: I'd prefer to be more precise here: `children_count_check` or `check_children_count`

marcandre (commented):
Thanks for the PR. I'm fine with replacing `arity` with `children_count`, but not just `children`.

lib/rubocop/ast/traversal.rb
Earlopain: Changed, the second one looks better to me

lib/rubocop/ast/traversal.rb
Earlopain: I renamed to `children_count_check_code` and moved the env variable handling there too so I don't have to think of another name for the local variable that used to hold the code

marcandre:
Thank you ❤️

PR #368, opened 2025-03-20
Remove a bit of backwards compat code

Earlopain:


Earlopain:
CI fails because of new unexpected node type: https://github.com/rubocop/rubocop-ast/pull/365

PR #369, opened 2025-03-20
Stop using `Parser::CurrentRuby`

Earlopain:
It emits a deprecation warning on Ruby 3.4 and won't support new versions. Instead:
* In tests use a fixed version that can be changed if needed
* Delegate to `ProcessedSource` and `RUBY_VERSION` to do the right thing. Eventually this will use the prism translation automatically

### DIFF
--- a/lib/rubocop/ast/node_pattern/compiler/debug.rb
+++ b/lib/rubocop/ast/node_pattern/compiler/debug.rb
@@ -109,14 +109,7 @@ module RuboCop
             private
 
             def ruby_ast(ruby)
-              buffer = ::Parser::Source::Buffer.new('(ruby)', source: ruby)
-              ruby_parser.parse(buffer)
-            end
-
-            def ruby_parser
-              require 'parser/current'
-              builder = ::RuboCop::AST::Builder.new
-              ::Parser::CurrentRuby.new(builder)
+              ProcessedSource.new(ruby, RUBY_VERSION.to_f, '(ruby)').ast
             end
           end
 
--- a/spec/rubocop/ast/node_pattern_spec.rb
+++ b/spec/rubocop/ast/node_pattern_spec.rb
@@ -1,16 +1,13 @@
 # frozen_string_literal: true
 
-require 'parser/current'
-
 RSpec.describe RuboCop::AST::NodePattern do
   include RuboCop::AST::Sexp
 
   def parse(code)
-    buffer = Parser::Source::Buffer.new('(string)', 1, source: code)
-    builder = RuboCop::AST::Builder.new
-    Parser::CurrentRuby.new(builder).parse(buffer)
+    RuboCop::AST::ProcessedSource.new(code, ruby_version, '(string)').ast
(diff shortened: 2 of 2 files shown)


lib/rubocop/ast/node_pattern/compiler/debug.rb
```diff
@@ -109,14 +109,7 @@ def test(ruby, trace: self.class::Compiler::Trace.new)
...
             def ruby_ast(ruby)
-              buffer = ::Parser::Source::Buffer.new('(ruby)', source: ruby)
-              ruby_parser.parse(buffer)
-            end
-
-            def ruby_parser
-              require 'parser/current'
-              builder = ::RuboCop::AST::Builder.new
-              ::Parser::CurrentRuby.new(builder)
+              ProcessedSource.new(ruby, RUBY_VERSION.to_f, '(ruby)').ast
```
Earlopain, line 112: I've once again been made aware that doing string.to_f is _not_ a great idea. Consider Ruby 3.10 for example, which would become 3.1.

But, since `ProcessedSource` accepts a float only, there's not much to do except hoping that Ruby 3.10 will never be released.

koic: How about applying the same patch as in https://github.com/ruby/prism/pull/3511?

Earlopain: That patch is good since it only ever is a string there. If I were to do that here, I'd end up with a string like "3.3", or "3.10", but rubocop-ast doesn't recognize that and I'd have to convert it into a float again.

Earlopain: It's a potential problem 6 years in the future. For example, the parsers also use simple integers like 33 or 27 to signify their version, which would be 310 for Ruby 3.1, messing a bunch of assumptions up.

koic: I see. Since I think leaving a comment about that risk is a reasonable choice, can you add one?

The following is off-topic. While the future is uncertain, AFAIK, Ruby 3.10 is unlikely to happen. There's a theory that if "Namespace" is introduced to Ruby in the near future, it will mark the arrival of Ruby 4. And that future may not be too far off.
https://bugs.ruby-lang.org/issues/19744

bbatsov: I agree it's better to focus on the immediate problems we need to solve and not worry too much about this distant (and quite uncertain) future.

bbatsov (approved):

lib/rubocop/ast/node_pattern/compiler/debug.rb
Earlopain: I would still have added a comment later :sweat_smile:. Anyways, this one is hardly a problem, I just saw it in tests all the time. This class if pretty much only used for the node pattern website I believe

PR #370, opened 2025-03-20
Support `Prism::Translation::Parser35` for Ruby 3.5 parser

Earlopain:
One step further towards https://github.com/rubocop/rubocop/issues/13617

In RuboCop, the parser engine in the default config can be set to `default`, and  with Ruby 3.5, RuboCop will then automatically choose prism to do its job.

More discussion and considerations in https://github.com/rubocop/rubocop-ast/pull/371

### DIFF
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -109,9 +109,13 @@ for the `:prism_result` keyword argument. Otherwise, the source code is parsed.
 
 [source,ruby]
 ----
-# Using the Parser gem with `prism_result: nil` is the default, meaning the source code is parsed.
+# Using the Parser gem with `prism_result: nil` is the default, meaning the source code is parsed until `ruby_version` is 3.4.
 ProcessedSource.new(@options[:stdin], ruby_version, file, parser_engine: :parser_prism, prism_result: :prism_result)
 ----
+
+IMPORTANT: The Parser gem supports syntax up to Ruby 3.3, but it does not support syntax in Ruby 3.4,
+such as `it` block parameters. Additionally, there are no plans to support Ruby 3.5 or later.
+For Ruby 3.5 and later, `parser_engine: parser_prism` is chosen automatically.
 
 This is an experimental feature. If you encounter any incompatibilities between
 Prism and the Parser gem, please check the following URL:
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -35,25 +35,21 @@ module RuboCop
       INVALID_LEVELS = %i[error fatal].freeze
       private_constant :INVALID_LEVELS
 
-      PARSER_ENGINES = %i[parser_whitequark parser_prism].freeze
+      PARSER_ENGINES = %i[default parser_whitequark parser_prism].freeze
       private_constant :PARSER_ENGINES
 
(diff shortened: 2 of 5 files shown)


lib/rubocop/ast/processed_source.rb
```diff
@@ -35,25 +35,21 @@ class ProcessedSource # rubocop:disable Metrics/ClassLength
       INVALID_LEVELS = %i[error fatal].freeze
       private_constant :INVALID_LEVELS
 
-      PARSER_ENGINES = %i[parser_whitequark parser_prism].freeze
+      PARSER_ENGINES = %i[automatic parser_whitequark parser_prism].freeze
```
koic, line 38: The term "automatic" might not be the most suitable. "default" might be more preferable.

```suggestion
      PARSER_ENGINES = %i[default parser_whitequark parser_prism].freeze
```

Earlopain: I'm not sure if `default` is actually better but I also don't like `automatic` particularly.

In the original version, I opted to simply use `nil` to signify this instead. But that has backwards compatibility concerns, where it might choose prism when it previously wouldn't when the user doesn't pass `parser_engine` to the method at all.

I changed it to `:default`.

koic: Yeah, I'm not sure if `default` is the most suitable choice, but it's a simple term. It's understood that this is a parameter for compatibility and it would be nice if a better name were suggested :-)

Earlopain: Yup, most users won't have to configure this. So I wouldn't think _too_ much about it.

docs/modules/ROOT/pages/index.adoc
```diff
@@ -109,10 +109,14 @@ for the `:prism_result` keyword argument. Otherwise, the source code is parsed.
 
 [source,ruby]
 ----
-# Using the Parser gem with `prism_result: nil` is the default, meaning the source code is parsed.
+# Using the Parser gem with `prism_result: nil` is the default, meaning the source code is parsed  until `ruby_version` is 3.4.
```
koic, line 112:
```suggestion
# Using the Parser gem with `prism_result: nil` is the default, meaning the source code is parsed until `ruby_version` is 3.4.
```

lib/rubocop/ast/processed_source.rb
```diff
@@ -397,6 +397,31 @@ def create_parser(ruby_version, parser_engine, prism_result)
...
+          best_parser_engine(ruby_version)
+        else
+          parser_engine
+        end
+      end
+
+      # The Parser gem does not support Ruby 3.5 or later.
+      # It is also not fully compatible with Ruby 3.4 but for
+      # now respects using parser for backwards compatibility.
+      def best_parser_engine(ruby_version)
```
koic, line 417: Can you rename it to the following?

```suggestion
      def default_parser_engine(ruby_version)
```

lib/rubocop/ast/processed_source.rb
bbatsov: Agreed.

bbatsov:
🚀

Earlopain:
Thanks for working with me on this @koic

Earlopain:
I openend https://github.com/rubocop/rubocop/pull/14025 since I already had half of it in https://github.com/rubocop/rubocop/pull/14010 to make it work. Hope that's ok.

bbatsov:
Of course!

PR #371, opened 2025-03-21
Support `Prism::Translation::Parser35` for Ruby 3.5 parser

koic:
## Summary

`Prism::Translation::Parser35` has been started development for Ruby 3.5 (edge Ruby): https://github.com/ruby/prism/pull/3346

At present, there is no timeline for Ruby 3.5 support in the Parser gem: https://github.com/whitequark/parser/issues/1046

Given this, support for Ruby 3.5 will first be introduced in `Prism::Translation::Parser`. This can also serve as a trigger to facilitate user migration from `ParserEngine: parser_whitequark` to `ParserEngine: parser_prism`.

As for Ruby 3.4, `Prism::Translation::Parser` is working on supporting the `it` syntax in https://github.com/ruby/prism/pull/3481, but the Parser gem has not yet supported for it. However, whether to deprecate Ruby 3.4 in the Parser gem will be considered separately from this PR.

Since Ruby 3.5 is a development version with minimal impact on users, while Ruby 3.4 is a stable release, they will be evaluated separately.

## Additional Information

First, the default `parser_engine` for Ruby 3.5 is set to `parser_prism`. This default aligns with the current state of support, as the Parser gem is not expected to support Ruby 3.5.

The more challenging decision is how to handle the default for Ruby 3.4. Ruby 3.4 will likely serve as a transition period between the Parser gem and Prism. While it is possible to set the default `parser_engine` to `parser_prism` for Ruby 3.4 as well, considering the potential unexpected incompatibilities in `Prism::Translation::Parser`, the default remains `parser_whitequark`.

In any case, the primary use case is RuboCop, and what matters most is which value RuboCop chooses to pass. In other words, the key decision lies with RuboCop.

### DIFF
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -109,9 +109,13 @@ for the `:prism_result` keyword argument. Otherwise, the source code is parsed.
 
 [source,ruby]
 ----
-# Using the Parser gem with `prism_result: nil` is the default, meaning the source code is parsed.
+# Using the Parser gem with `prism_result: nil` is the default, meaning the source code is parsed until `ruby_version` is 3.4.
 ProcessedSource.new(@options[:stdin], ruby_version, file, parser_engine: :parser_prism, prism_result: :prism_result)
 ----
+
+The Parser gem supports syntax up to Ruby 3.3, but it does not support syntax in Ruby 3.4,
+such as `it` block parameters. Additionally, there are no plans to support Ruby 3.5 or later.
+From Ruby 3.5, only `parser_engine: parser_prism` may be supported by default.
 
 This is an experimental feature. If you encounter any incompatibilities between
 Prism and the Parser gem, please check the following URL:
--- a/changelog/new_support_ruby35.md
+++ b/changelog/new_support_ruby35.md
@@ -1,0 +1,1 @@
+* [#371](https://github.com/rubocop/rubocop-ast/pull/371):Support `Prism::Translation::Parser35` for Ruby 3.5 parser (experimental). ([@koic][])
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -46,13 +46,35 @@ module RuboCop
         new(file, ruby_version, path, parser_engine: parser_engine)
       end
 
+      # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       def initialize(
         source, ruby_version, path = nil, parser_engine: :parser_whitequark, prism_result: nil
       )
-        parser_engine = parser_engine.to_sym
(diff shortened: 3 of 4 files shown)


docs/modules/ROOT/pages/index.adoc
```diff
@@ -109,10 +109,14 @@ for the `:prism_result` keyword argument. Otherwise, the source code is parsed.
 
 [source,ruby]
 ----
-# Using the Parser gem with `prism_result: nil` is the default, meaning the source code is parsed.
+# Using the Parser gem with `prism_result: nil` is the default, meaning the source code is parsed until `ruby_version` is 3.4.
 ProcessedSource.new(@options[:stdin], ruby_version, file, parser_engine: :parser_prism, prism_result: :prism_result)
 ----
 
+The Parser gem supports syntax up to Ruby 3.3, but it does not support syntax in Ruby 3.4,
```
bbatsov, line 116: Might be good to add `IMPORTANT:` here.

bbatsov (approved):

Earlopain:
I don't think this is the best solution, the user should not see such a warning. You have given me an idea:

* Modify my PR in https://github.com/rubocop/rubocop-ast/pull/370 and simply tweak the version number it chooses prism from to start from 3.5
* You simply add the new parser to the list here
* RuboCop changes its default `TargetRubyVersion` to `automatic` like in https://github.com/rubocop/rubocop/pull/14010

It will then choose prism automatically for Ruby 3.5 only and there is nothing the user must do. Also, in the future rubocop-ast could simply choose to automatically switch to prism for ruby 3.4 as well and it would only be a single change. There would be no breaking change for how it currently does things.

What do you think?

Earlopain:
I modified my PR to show what I meant and added you as co-author.

koic:
@Earlopain Yeah, regarding #370, if there are no incompatibilities for existing users, it seems reasonable to collaborate based on that approach.
First, it's preferable to maintain the current state without a default dependency on Prism. Adding such a default dependency can be considered at a later stage.
As for `ParserEngine: automatic`, the understanding is that it maintains compatibility. Is that correct?

Earlopain:
Yes, since it is a completely new option there is nothing that can be broken. Since I made it to only choose prism on 3.5, there are also no issues for users that currently don't explicitly pass `parser_engine`, because it would simply continue to use `parser`. Only when opted in to 3.5 would it then error when prism is not present, same as in your PR here.

I guess the dependency on prism can still be delayed to a later date, but it has to happen eventually. Let me check how it would look to keep the dependency optional.

koic:
Nice. It will eventually be necessary due to the dependency on `Prism::Translation::Parser`. However, it would be better to slightly delay the timing to make it easier to isolate user impact and potential issues.

koic:
I will close this PR because it has been integrated into #370.

PR #372, opened 2025-03-21
Remove prism skips in tests

Earlopain:
These all just pass now.

The traversal spec is a bit special. With `parser`, Ruby 2.7 is used but `prism` can only do 3.3. So some code samples are now invalid syntax on newer rubies. I simply removed those.


marcandre:
Thanks!

PR #375, opened 2025-03-26
Use a fixed version of racc

Earlopain:
The gem contains files generated via it and I noticed that the most recent release used 1.8.1, while the one before was created with 1.5.0

There are some differences in the generated files. It seems best to always use an explicit version, similar to how it's done in the `parser` gem. https://github.com/whitequark/parser/pulls?q=is%3Apr+racc

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ gemspec
 gem 'bump', require: false
 gem 'bundler', '>= 1.15.0', '< 3.0'
 gem 'oedipus_lex', '>= 2.6.0', require: false
-gem 'racc'
+gem 'racc', '1.8.1'
 gem 'rake', '~> 13.0'
 gem 'rspec', '~> 3.7'
 gem 'simplecov', '~> 0.20'


bbatsov:
Shouldn't we relax the patch version at least?

Earlopain:
I have it at an exact version since `parser` also does that. Generally, this makes sense to me? If a future version contains some fixes then those fixes should also be part of the next release, regardless who who pushed the gem.

There are barely any releases of `racc` so the maintenance burden is pretty low. @koic would know, do you have an opinion on this PR?

marcandre:
I prefer keeping it either entirely relaxed (with the thinking that there's nothing to indicate that a version 2 would necessarily break this gem), or if not fixed to a particular major version.

koic:
Based on the investigation, this goes back as far as https://github.com/whitequark/parser/commit/af6cc2b70b41d8cde8e6c782efc7dd5489b38987. In practice, since I started following the project, I don’t recall any issues caused by Racc upgrades. So, there doesn't seem to be a need for RuboCop AST (and RuboCop) to specify a version constraint on Racc.

As a side note, if Racc were to support new syntax in `parse.y` in the future similar to Lrama parser generator and that syntax were used in a repository-managed `parse.y` file (e.g., `lib/rubocop/ast/node_pattern/parse.y`), then it would become necessary to guarantee a minimum version. However, there is currently no sign of that happening.

Earlopain:
Just a note, 1.5.0 was released 5 years ago. I have no real concerns about incompatibilies from that, but flip-flopping so majorly between racc versions depending on who published the gem feels wrong.

bbatsov:
No, argument from me. That's why I suggested just narrowing this to `1.8`. At any rate - I don't have a strong preference here and I'll defer to @koic and @marcandre.

PR #376, opened 2025-03-26
Manually parse node patterns

kddnewton:
Currently, both the lexer and parser are generated in order to parse node patterns. Instead, we could parse them manually and avoid having to generate them and the complexity that adds.

---

I'm not sure if you want this or not. Certainly the current lexer/parser combination is working just fine. I only built this because I've been looking at transpiling the node matcher AST to something that can match against Prism nodes, and I don't want to take on the additional two dependencies of the generators.

I haven't benchmarked this at all, because I highly doubt this is a hotspot literally anywhere, but I would assume it is very slightly faster.

There are many other places I think this could be improved (by changing some of the token names, inlining some more of the parsing functions, etc) but I wanted to keep all of the tests the same and I wasn't exactly sure what was public API and what wasn't (for example, it's clear the Lexer can be extended). If you want this to go further, that's certainly possible.

Either way, I'm happy to have this closed, I'm going to copy it over to Prism to use it for my own purposes. Just wanted to offer this in case it's useful!


marcandre:
> Currently, both the lexer and parser are generated in order to parse node patterns.

Correct. They are generated from `.rex` and `.y` files which describe succinctly and clearly how to parse them.

> Instead, we could parse them manually and avoid having to generate them and the complexity that adds.

Only the developer has to generate them, and only when either of the `.rex` and `.y` file is changed.

I'm not sure which complexity this adds, but I am sure you underestimate the complexity of making any change to the parsing of node patterns from the generated file. It's a bit like proposing to throw away the `.c` code and modifying directly the assembly code if further changes are desired, in order to avoid the need for a C compiler and learning the C language...

> [...] I don't want to take on the additional two dependencies of the generators.

It is your personal choice, but I strongly encourage you to instead learn `.y` and `.rex` formats, and embracing these two dependencies (for development only).

kddnewton:
Thank you for considering the change. To address a couple of your points:

"describe succinctly" is very subjective. I understand it, you understand it, new developers will not understand it. It is a separate language, requiring additional study.

The complexity it adds is to the build system, as well as taking on the security of various dependencies.

Finally, I assure you, I am well versed in the `.y` and `.rex` formats.

Regardless, I understand that you do not want to incorporate this change.

PR #377, opened 2025-04-01
Add support `RuboCop::AST::Node#any_def_type?` method

koic:
This PR adds support for `RuboCop::AST::Node#any_def_type?` method, which abstracts over the `def` and `defs` node types.


Earlopain:
👍 I also think this makes sense. I though about adding this myself in the past but couldn't find a good name. `any_def` looks good to me.

There are many places in RuboCop than can make use of this, from just a basic search for `rubocop`:
grafik

bbatsov (approved):
Good idea!

PR #378, opened 2025-04-11
Fix `flip-flop` operator possible children nodes count

viralpraxis:
One of the "flip-flop"'s operator nodes might be `nil`:

```bash
$ ruby-parse --34 -e '!(...(a == 1))'
(send
  (begin
    (eflipflop nil
      (begin
        (send
          (send nil :a) :==
          (int 1))))) :!)
```

```bash
$ docker run -it --rm rubylang/all-ruby env ALL_RUBY_SINCE=ruby-2.7.0 ./all-ruby -e '!(..(a==1))'
ruby-2.7.0
...
ruby-3.4.1
```


marcandre:
Good catch, thanks!

marcandre:
Released as 1.44.1

For future contributions: please add a changelog entry using `rake changelog:fix` or variation (I created one for this PR).

Thanks again

PR #379, opened 2025-05-02
Support `RuboCop::AST::ComplexNode`

koic:
This PR supports `RuboCop::AST::ComplexNode`.

The `complex` type is categorized as a `numeric` type, so it should be handled transparently alongside `int`, `float`, and `rational`.

```ruby
int: :numeric,
float: :numeric,
rational: :numeric,
complex: :numeric,
```

https://github.com/rubocop/rubocop-ast/blob/v1.44.1/lib/rubocop/ast/node.rb#L106-L109

This eliminates the need for redundant branching, such as in https://github.com/rubocop/rubocop/pull/14146.


marcandre:
Oh, sorry about missing the notifications for this. Thank you!

marcandre:
Released as v1.46.0

PR #381, opened 2025-06-03
Support `AST::Node#any_match_pattern_type?` method

koic:
This PR adds `AST::Node#any_match_pattern_type?`.

This method can be used to simplify multiple places where `node.type?(:match_pattern, :match_pattern_p)` is used, such as in the changes made in https://github.com/rubocop/rubocop/pull/14224.


marcandre:
Nice, thank you ❤️

Released as 1.45.0

PR #382, opened 2025-06-06
Fix an error when using Parser gem with `prism_result`

koic:
This PR fixes an error when using Parser gem with `prism_result`. https://github.com/rubocop/rubocop/actions/runs/15497886582/job/43638843391?pr=14251

Since Ruby LSP 0.24 or later, `prism_result` is now passed as a `Prism::ParseLexResult`. As a result, the following build error occurs in RuboCop for the Ruby LSP add-on.

The processing that uses `prism_result` assumes Prism and cannot be performed when using the Parser gem.


marcandre:
Thanks! Released as 1.45.1

PR #383, opened 2025-07-29
Remove CodeClimate from CI

Earlopain:
CodeClimate as a coverage service has been discontinued.

https://codeclimate.com/blog/code-climate-quality-is-now-qlty-software
https://docs.qlty.sh/migration/coverage
> The Code Climate API will be disabled on July 18th, 2025.

In RuboCop, the coverage recording was simply removed instead of setting up an alternative. Can we do the same here? Also see https://github.com/rubocop/rubocop/issues/14392


marcandre:
Thanks! 👍

PR #385, opened 2025-08-25
Sync spellcheck from rubocop repo

Earlopain:
Does two things:
* Removes misspell since the project has been archived and codespell exists
* Runs codespell locally with `bundle exec rake`, if available

Commits contain references to the original PRs from rubocop


bbatsov:
Thanks!

PR #386, opened 2025-09-02
Allow to match against the empty string

Earlopain:
You can work around this by writing `(str empty?)` instead but I feel like this should just work

cc @zopolis4


marcandre:
Released in 1.47.1

Earlopain:
Ty

PR #387, opened 2025-09-19
Add `Node#any_sym_type?` to match `sym` and `dsym` types

dvandersluis:
Following #328, adds `any_sym_type?` to match `sym` and `dsym` nodes.


marcandre:
Released as 1.47.0, thanks ❤️

PR #388, opened 2025-09-19
Fix build error in CI

dvandersluis:
The "very old specs" check has been failing [since August](https://github.com/rubocop/rubocop-ast/actions/runs/17205345094).


dvandersluis:
Oh, interesting. Thanks for the insight @Earlopain. I changed it back to the older RuboCop, but pinned `prism` below 1.5.0 like in your PR, which allows the check to pass again.

marcandre:
@Earlopain and @dvandersluis you are so awesome! I was dreading having to figure this out myself 😅.

Thanks so much! ❤️.

dvandersluis:
Happy to help @marcandre! Thanks for all you do!

PR #401, opened 2026-03-14
Minor performance improvements

bbatsov:
## Summary

* **Fix `ProcessedSource` `force_encoding` for frozen string inputs** — `(+source).force_encoding(...)` was creating a mutable copy but discarding the result, so the encoding fix was silently a no-op when a frozen string was passed. Now assigns the result back.
* **Avoid intermediate array allocations in `HashNode#each_key`/`#each_value`** — the old implementation called `pairs.map(&:key).each(&block)`, allocating a full intermediate array on every call. Now yields directly from `each_child_node`.

## Benchmark results

```
--- HashNode#each_key ---
  each_key (current):  1055691.9 i/s
  each_key (old):       625038.9 i/s - 1.69x slower

--- HashNode#each_value ---
  each_value (current):  1076484.9 i/s
  each_value (old):       591610.7 i/s - 1.82x slower

--- Allocation counts (100 calls) ---
  each_key: old=1301 allocs, new=600 allocs
```

### DIFF
--- a/changelog/fix_avoid_intermediate_array_allocations_in.md
+++ b/changelog/fix_avoid_intermediate_array_allocations_in.md
@@ -1,0 +1,1 @@
+- [#401](https://github.com/rubocop/rubocop-ast/pull/401): Avoid intermediate array allocations in HashNode#each_key/each_value. ([@bbastov][])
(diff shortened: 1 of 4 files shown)


marcandre:
Nice 👍
Adding changelog entries...

PR #404, opened 2026-07-04
Build prism translation tokens lazily

bbatsov:
Converting prism tokens to the parser gem's format is a significant part of the translation cost (about 40% in my measurements, `#parse` vs `#tokenize` over a 450-file corpus), and not every caller of `ProcessedSource` looks at the tokens at all.

This defers the token conversion until first access. The AST and comments are still built eagerly from the single `parse_lex` call, whose result is retained, so nothing gets parsed twice and the tokens come out byte-identical to the eager path (including for invalid syntax and encoding errors). The whitequark path is unchanged, since there the tokenize-vs-parse difference is around 1%.

One implementation note: the lazy parsers are memoized subclasses of the translation parsers rather than per-instance `extend`s. My first version extended each parser instance, and the fresh singleton class per file defeated method caches in the translation internals, which cost about 5% when tokens were used. With real subclasses it's parity.

Numbers: `ProcessedSource` creation is ~32% faster with the prism engine when tokens are never accessed, unchanged when they are. On a real run, `rubocop --only Style/Not` with `ParserEngine: parser_prism` over rubocop's `lib/rubocop/cop` goes from ~2.1s to ~1.8s. Default full runs are unaffected (Layout cops demand tokens on virtually every file); the win is for `--only` runs, plugins and API consumers like Ruby LSP.

Both `rake spec` and `rake prism_spec` pass here, and RuboCop's entire `prism_spec` suite passes against this branch. Longer term I'd like to propose a supported API for this on the prism side, since `tokenize_deferred` currently reuses the translation parser's private helpers.


marcandre:
Makes sense, thanks!

PR #405, opened 2026-07-04
Use Prism by default for Ruby 3.3 analysis

bbatsov:
Currently `parser_engine: :default` resolves to Prism only for target Ruby 3.4+, and 3.3 still goes through the Parser gem. That boundary made sense when the translation layer was fresh, but Prism parses as 3.3 just fine (`Prism::Translation::Parser33`), it's about 2x faster than whitequark on the ProcessedSource level, and the compatibility issue tail has been quiet for a while. Both rubocop-ast suites and RuboCop's entire spec suite pass with the flip.

Anyone who needs the old behavior can still set `parser_engine: :parser_whitequark` explicitly, and RuboCop exposes that via `ParserEngine`.

Related: #404, and longer term I'd like us to move to a native prism builder so the translation layer can be dropped entirely.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -26,6 +26,10 @@ InternalAffairs/LocationExpression:
   Enabled: false
 
 # It cannot be replaced with suggested methods defined by RuboCop AST itself.
+InternalAffairs/LocationLineEqualityComparison:
+  Enabled: false
+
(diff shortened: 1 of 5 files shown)


marcandre:
LGTM, thanks 👍

PR #406, opened 2026-07-04
Fix offenses flagged by RuboCop master

bbatsov:
CI's Coding Style job runs against RuboCop master, which recently gained `InternalAffairs/LocationLineEqualityComparison` and tightened redundant-disable detection, so every PR is currently red on that job. The new cop suggests `same_line?`, which is a RuboCop-side helper not available here, so it gets disabled like the other InternalAffairs cops whose suggestions don't apply to rubocop-ast. The `Style/OneClassPerFile` disable in spec_helper is simply no longer needed.

Same commit is already cherry-picked into #404 and #405; merging this first unblocks CI for everything else.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -26,6 +26,10 @@ InternalAffairs/LocationExpression:
   Enabled: false
 
 # It cannot be replaced with suggested methods defined by RuboCop AST itself.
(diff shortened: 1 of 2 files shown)


marcandre:
Thanks ❤️

PR #407, opened 2026-07-04
Fix the enumerator form of the deprecated each_* branch methods

bbatsov:
The blockless form of `each_when`, `each_in_pattern`, `each_condition` and `each_branch` called `to_enum` on the branches array instead of the node, so the returned enumerator raised `NoMethodError` the moment you iterated it. Broken since 2020 apparently, which tells you how popular these deprecated methods are, but they're still shipped API (and RuboCop's `Lint/Void` still uses the block form), so they should work until 2.0 removes them.

The existing specs only asserted `be_a(Enumerator)` without ever iterating, which is exactly how this hid; they now consume the enumerator too.

Includes the same lint-fix commit as #406 so CI is green; merging #406 first makes it disappear here.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -26,6 +26,10 @@ InternalAffairs/LocationExpression:
   Enabled: false
 
 # It cannot be replaced with suggested methods defined by RuboCop AST itself.
+InternalAffairs/LocationLineEqualityComparison:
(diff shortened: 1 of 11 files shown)


marcandre:
Pretty funny 😅. Thanks.

PR #408, opened 2026-07-04
Small ProcessedSource fixes

bbatsov:
Three small independent fixes:

* `checksum` recomputed the SHA1 on every call even though the source never changes after construction; RuboCop calls it once per autocorrect iteration. Now memoized.
* `sorted_tokens` allocated a tuple per token to sort a list that's already ordered for everything except heredoc-argument files. It now checks order first (allocation-free) and only sorts when needed.
* `preceding_line` for a token on line 1 did `lines[-1]` and silently returned the last line of the file. It returns nil now.


marcandre:
Looks good 👍
Resolving conflict...

PR #409, opened 2026-07-04
Use identity comparison in chained? and block_literal?

bbatsov:
`chained?` and `block_literal?` compared with `eql?`, which does deep structural equality, so in `foo.bar(foo)` the argument node claimed to be chained (it deep-equals the receiver), and in `foo { foo }` the block body claimed to be a block literal. Since these predicates ask "am I this specific child of my parent", `equal?` is what's meant, and it's also free instead of a recursive tree comparison on a hot path.

Both regression specs fail against the old code. I checked all callers in rubocop-ast and rubocop; none relied on the structural-equality behavior.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -26,6 +26,10 @@ InternalAffairs/LocationExpression:
   Enabled: false
 
 # It cannot be replaced with suggested methods defined by RuboCop AST itself.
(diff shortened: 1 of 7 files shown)


marcandre:
Good catch 👍

PR #410, opened 2026-07-04
Reduce allocations in hot node predicates

bbatsov:
Two allocation fixes on per-send hot paths:

* The method-name predicates (`assignment_method?`, `enumerator_method?`, `predicate_method?`, `bang_method?`, `camel_case_method?`) converted the method name Symbol to a String on every call. `Symbol#start_with?`/`#end_with?` exist since Ruby 2.7, which is our floor, so the conversions are pointless. `camel_case_method?` now uses `match?` and returns a real boolean instead of `0`/nil; all callers use it in boolean context only.
* `keyword?` sliced the node's source out of the buffer for every node just to check for `__FILE__`/`__LINE__`/`__ENCODING__`, which can only ever appear as a handful of node types. A type check now guards the slice; behavior is bit-for-bit identical, including the quirk that symbol/string literals spelled like special keywords count (verified empirically against the old implementation across the odd cases: `%i[__FILE__]`, dstr parts, hash labels).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -26,6 +26,10 @@ InternalAffairs/LocationExpression:
   Enabled: false
 
 # It cannot be replaced with suggested methods defined by RuboCop AST itself.
+InternalAffairs/LocationLineEqualityComparison:
+  Enabled: false
+
+# It cannot be replaced with suggested methods defined by RuboCop AST itself.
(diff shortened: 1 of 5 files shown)


marcandre:
Nice free optimizations 🎉

PR #411, opened 2026-07-04
Don't allocate an array per node visited in filtered traversals

bbatsov:
Filtered traversals (`each_descendant(:send)`, `each_node`, `each_ancestor`, `each_child_node`) called `child.type?(*types)` per visited node, which re-splats the types array into a fresh Array every time, and re-tested `types.empty?` per node although it's invariant for the whole walk. These traversals back every `def_node_search` and hundreds of direct call sites in cops, so it adds up.

This adds `Node#type_in?(types)`, a non-splatting variant of `type?` (which now delegates to it), and splits the descendant/ancestor visits into filtered and unfiltered variants chosen once per walk. `visit_descendants` keeps its old signature as a shim since it's protected API.

Measured on a ~1300-node tree: a filtered walk goes from 1336 allocations to 1 and about 13% faster wall time; the allocation elimination also cuts GC pressure across full runs. Unfiltered walks are unchanged.

### DIFF
--- a/lib/rubocop/ast/node/mixin/descendence.rb
+++ b/lib/rubocop/ast/node/mixin/descendence.rb
@@ -60,11 +60,7 @@ module RuboCop
       def each_descendant(*types, &block)
         return to_enum(__method__, *types) unless block
 
-        if types.empty?
-          visit_all_descendants(&block)
-        else
-          visit_descendants_of_types(types, &block)
-        end
+        visit_descendants(types, &block)
 
         self
       end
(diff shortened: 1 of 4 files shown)


lib/rubocop/ast/node/mixin/descendence.rb
```diff
@@ -60,7 +60,11 @@ def child_nodes
       def each_descendant(*types, &block)
         return to_enum(__method__, *types) unless block
 
-        visit_descendants(types, &block)
+        if types.empty?
+          visit_all_descendants(&block)
+        else
+          visit_descendants_of_types(types, &block)
+        end
```
marcandre: Any reason for this change? I don't see why we'd duplicate `visit_descendants`'s implementation here.

marcandre:
Nice optimization. Once my minor comment is addressed, I'll merge and make a release 👍

lib/rubocop/ast/node/mixin/descendence.rb
bbatsov: No good reason; it saved one method call per walk, which is noise. Simplified as suggested, and rebased on master to drop the lint-fix commit that already landed via #406.

marcandre:
Released, along with all other optimizations in 1.50.0 🚀

PR #412, opened 2026-07-04
Use identity comparison in argument?

bbatsov:
Same bug class as #409, this time in `argument?`: `parent.arguments.include?(self)` uses deep structural equality, so in `foo.bar(foo)` the *receiver* claims to be an argument because it deep-equals the actual argument. The predicate means "am I one of these specific child nodes", so it now scans by identity, which also replaces recursive tree comparisons with pointer checks on a per-node hot path.

There were no `argument?` specs at all; the new ones include the receiver regression (fails against the old code).


marcandre:
Good catch 👍

PR #413, opened 2026-07-04
Add same_line?, begins_its_line?, any_descendant? and first_part_of_call_chain

bbatsov:
These four helpers have lived in RuboCop's `Cop::Util` although they're pure AST/range logic, which is why our own `.rubocop.yml` had to disable `InternalAffairs/LocationLineEqualityComparison` with the note that the suggested method doesn't exist in rubocop-ast. Moving them down lets RuboCop and every extension use them, and the cop is re-enabled here since `loc.begin.same_line?(loc.end)` is now expressible (used by `BlockNode#single_line?` and `ConditionalNode#single_line_condition?`).

Placement notes: `same_line?` and `begins_its_line?` go on `Ext::Range` rather than `Node`, deliberately, because `HashElementNode#same_line?` already exists with different semantics (multiline overlap counts as same line) and a `Node#same_line?` would silently diverge for pair nodes. `same_line?` accepts ranges, nodes, tokens, and comments, mirroring `Cop::Util#same_line?`'s duck-typing. `any_descendant?` respects its type arguments in the blockless form (the `Cop::Util` version ignores them there, but nothing calls it that way). `first_part_of_call_chain` keeps the `Cop::Util` behavior exactly, including returning nil for receiverless chains.

Once this is released I'll send the RuboCop-side PR making `Cop::Util` delegate to these.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -26,10 +26,6 @@ InternalAffairs/LocationExpression:
   Enabled: false
 
 # It cannot be replaced with suggested methods defined by RuboCop AST itself.
-InternalAffairs/LocationLineEqualityComparison:
-  Enabled: false
-
-# It cannot be replaced with suggested methods defined by RuboCop AST itself.
 InternalAffairs/MethodNameEndWith:
   Enabled: false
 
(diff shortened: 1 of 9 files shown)


marcandre:
Looks good, thanks

PR #414, opened 2026-07-04
Compile node patterns lazily

bbatsov:
`def_node_matcher`/`def_node_search` compile their pattern (a racc parse plus codegen, ~0.15ms and ~300 allocations each) at definition time. RuboCop defines ~540 patterns and extensions add hundreds more, so this is a meaningful chunk of boot, paid even for `rubocop --only Style/Foo` runs where almost none of the patterns are ever matched. Measured with RuboCop against this branch: `require "rubocop"` goes from ~370ms to ~270ms.

The macros now define a stub that compiles the pattern and replaces itself on first call. Design notes:

* The stub removes itself before the compiled method is defined, so nothing is emitted under `-w` (RuboCop's spec suite runs with `$VERBOSE` on and caught my first attempt).
* The compiled method is invoked through the owner's `instance_method`, so subclass overrides calling `super` work (spec'd).
* The definition location is captured eagerly in the macro and attached to the compiled method via the new `NodePattern#definition_location`, so backtraces and `caller`-sensitive specs still point at the cop source. The existing spec asserting error backtrace line numbers passes unchanged.
* One observable change: an invalid pattern raises `NodePattern::Invalid` at first call instead of at definition. `NodePattern.lazy_compilation = false` restores definition-time errors for anyone who wants load-time validation (e.g. a plugin's CI).

Both suites, self-lint, and RuboCop's entire spec suite (plus its prism variant) pass against this branch.

### DIFF
--- a/lib/rubocop/ast/node_pattern.rb
+++ b/lib/rubocop/ast/node_pattern.rb
@@ -26,6 +26,18 @@ module RuboCop
     class NodePattern
       class Invalid < StandardError; end
 
+      # When `true` (the default), methods defined with the `Macros` compile
+      # their pattern on first invocation instead of at definition time.
+      # Compiling the patterns is a significant part of loading a large body
+      # of cops, and most of them are never invoked in a given run.
+      # Note that with lazy compilation an invalid pattern raises
+      # `NodePattern::Invalid` when the method is first called, not when it
+      # is defined; set to `false` to get definition-time errors back.
+      class << self
+        attr_accessor :lazy_compilation
+      end
+      self.lazy_compilation = true
+
       # Helpers for defining methods based on a pattern string
       module Macros
         # Define a method which applies a pattern to an AST node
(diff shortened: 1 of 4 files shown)


lib/rubocop/ast/node_pattern.rb
```diff
@@ -26,6 +26,18 @@ module AST
...
+      # their pattern on first invocation instead of at definition time.
+      # Compiling the patterns is a significant part of loading a large body
+      # of cops, and most of them are never invoked in a given run.
+      # Note that with lazy compilation an invalid pattern raises
+      # `NodePattern::Invalid` when the method is first called, not when it
+      # is defined; set to `false` to get definition-time errors back.
+      class << self
+        attr_accessor :lazy_compilation
+      end
+      self.lazy_compilation = true
```
marcandre, line 39: Should we not set the default to `false`? For most other users of this gem, I think this is a better, less confusing default and it remains compatible.

marcandre: @bbatsov no rush, just making sure you saw this ☝️

PR #415, opened 2026-07-04
Make sorted_tokens allocation-free

bbatsov:
Follow-up to #408, which got two things wrong. The order check used `each_cons`, which allocates an array per pair, defeating its own purpose. And re-profiling showed the already-sorted fast path almost never fires on real code anyway: the lexer emits comment tokens out of band, so any file containing comments has out-of-order tokens and takes the sort. The doc comment claiming heredoc arguments are the only unsorted case was wrong too.

So this makes the actual sort cheap instead: position and index combined into one integer key keeps the sort stable without allocating a tuple per token (overflow would need a multi-GB file). The scan is index-based now as well. Measured on `node.rb` (comments and heredocs both): ~2700 allocations down to 20, identical output verified element-wise.

### DIFF
--- a/changelog/change_allocation_free_sorted_tokens.md
+++ b/changelog/change_allocation_free_sorted_tokens.md
@@ -1,0 +1,1 @@
+* [#415](https://github.com/rubocop/rubocop-ast/pull/415): Reduce allocations in `ProcessedSource#sorted_tokens` to almost none. ([@bbatsov][])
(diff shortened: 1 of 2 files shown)


marcandre:
Looks good.
I'm wondering if it wouldn't be worth it to cache the result. @bbatsov did you get a chance to measure that?

PR #416, opened 2026-07-04
Stop allocating in IfNode branch accessors

bbatsov:
Re-profiling after the last batch showed `IfNode#node_parts` as the biggest remaining source of derived-value allocations (~2.3% of a run's total): `condition`/`if_branch`/`else_branch` each built the three-element array per call, plus `unless?` sliced the keyword source out of the buffer every time, and `if` nodes are everywhere.

The accessors now index `children` directly (with the same `unless` swap), `unless?` compares the keyword length instead of allocating its source (a non-ternary `if` node's keyword can only be `if`, `elsif` or `unless`), and `node_parts` is expressed in terms of the accessors. `ConditionalNode#condition` reads `children[0]`, which holds for every includer since only `IfNode` overrides `node_parts` and even there the condition stays first.

Four thousand accessor calls now allocate once (the first `keyword` comparison); behavior verified identical against the old implementation across if/unless/ternary/modifier/elsif forms, and `IfNode` no longer appears in the allocation profile at all.

### DIFF
--- a/changelog/change_allocation_free_if_node_accessors.md
+++ b/changelog/change_allocation_free_if_node_accessors.md
@@ -1,0 +1,1 @@
+* [#416](https://github.com/rubocop/rubocop-ast/pull/416): Make `IfNode` branch accessors (`condition`, `if_branch`, `else_branch`, `unless?`) allocation-free. ([@bbatsov][])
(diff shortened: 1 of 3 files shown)


marcandre:
Thanks for the PR. I'll have a look asap, hopefully tomorrow.

marcandre:
Thanks 👍

PR #417, opened 2026-07-14
Fix `ProcessedSource` raising parser internal errors

### LINKED ISSUE #15457: Exception on invalid multibyte escape in ./ruby/test/ruby/test_regexp.rb
### Before submitting

- [x] I have searched the existing issues and this hasn't been reported yet.
- [x] I am running the latest stable version of RuboCop.

### Issue type

Crash/error

### Cop name

N/A

### Steps to reproduce the problem

/\c\xFF/ =~ "\c\xFF"

https://github.com/ruby/ruby/blob/master/test/ruby/test_regexp.rb#L799

### Expected behavior

No crash. Also, can Rubocop print which file and line number this happened in, to prevent human bisecting?

### Actual behavior

```
invalid multibyte escape: /\x9F/
/opt/homebrew/lib/ruby/gems/4.0.0/gems/parser-3.3.11.1/lib/parser/builders/default.rb:2276:in 'Regexp#initialize'

```

### RuboCop configuration

```yaml
/dev/null
```

### RuboCop version

```shell
System Version: macOS 26.5.1 (25F80)
Model Identifier: MacBookAir10,1
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +PRISM [arm64-darwin25]
rubocop     2026-07-14 84bcf77f7
```

koic:
The parser gem may fail on invalid sources with an exception other than `Parser::SyntaxError` and without a diagnostic. For example, `/\c\xFF/ =~ ""` raises `RegexpError` (invalid multibyte escape) from the static regexp validation, or `ArgumentError` (invalid byte sequence in UTF-8) from the lexer, depending on the parser engine and version.

`ProcessedSource#tokenize` only rescued `Parser::SyntaxError`, so these exceptions escaped from `ProcessedSource#parse` and crashed the caller. For RuboCop, this aborted the entire run without even telling which file was being inspected.

Such exceptions are now recorded as `parser_error`, the same mechanism already used for `EncodingError` and `Parser::UnknownEncodingInMagicComment` when setting the buffer source. RuboCop then reports them per file as a fatal `Lint/Syntax` offense and continues with the remaining files:

```console
$ rubocop example.rb
example.rb:1:1: F: Lint/Syntax: Invalid byte sequence in utf-8.
```

Fixes rubocop/rubocop#15457.

### DIFF
--- a/changelog/fix_an_error_for_parser_internal_exceptions.md
+++ b/changelog/fix_an_error_for_parser_internal_exceptions.md
@@ -1,0 +1,1 @@
+* [#417](https://github.com/rubocop/rubocop-ast/pull/417): Fix `ProcessedSource` raising parser internal errors (e.g. `RegexpError` for `/\c\xFF/ =~ ''`) instead of recording them as `parser_error`. ([@koic][])
--- a/lib/rubocop/ast/processed_source.rb
+++ b/lib/rubocop/ast/processed_source.rb
@@ -341,6 +341,13 @@ module RuboCop
           # All errors are in diagnostics. No need to handle exception.
           comments = []
           tokens = []
+        rescue StandardError => e
+          # The parser gem may fail on invalid sources with an exception other than `Parser::SyntaxError`
+          # and without a diagnostic, e.g. with `RegexpError` for `/\c\xFF/ =~ ""`. Keep the exception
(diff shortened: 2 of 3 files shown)


marcandre:
Thanks!

